PR #45, opened 2023-05-18
FMT read voltage tables to assign hit status

zieglerv:
Merge requested to verify the status tables for RGM review. .


reconstruction/fmt/src/main/java/org/jlab/rec/fmt/hit/Hit.java
```diff
@@ -292,6 +308,25 @@ public static List<Hit> fetchHits(DataEvent event, IndexedTable timecuts, Indexe
...
+            int i = strip_number -1;
+            // To represent the geometry we divide the barrel micromega disk into 3 regions according to the strip numbering system.
+            // Here i = strip_number -1;
+            // Region 1 is the region in the negative x part of reg region: the strips range is from   1 to 320  (   0 <= i < 320)
+            // Region 2 is the region in the negative y part of outer region: the strips range is from 321 to 512  ( 320 <= i < 512)
+            // Region 3 is the region in the positive x part of reg region: the strips range is from 513 to 832  ( 512 <= i < 832)
+            // Region 4 is the region in the positive y part of outer region: the strips range is from 833 to 1024 ( 832 <= i < 1024)
+            int reg =0; //2=outer region; true=1 region
+
+            if((i>=0 && i<320) || (i>=512 && i<832) ) {
```
raffaelladevita: would be better to move the numbers to the Constants class or use a CCDB table

reconstruction/fmt/src/main/java/org/jlab/service/fmt/FMTEngine.java
```diff
@@ -50,7 +50,10 @@ public boolean init() {
         String[] tables = new String[]{
             "/geometry/beam/position",
             "/calibration/mvt/fmt_time",
-            "/calibration/mvt/fmt_status"
+            "/calibration/mvt/fmt_status",
+            "/calibration/mvt/fmt_voltage_drift",
+            "/calibration/mvt/fmt_voltage_strip_inout",
```
raffaelladevita, line 55: is this table used? is it intended for the numbers mentioned above?

raffaelladevita (changes requested):

reconstruction/fmt/src/main/java/org/jlab/service/fmt/FMTEngine.java
zieglerv: Yes, this table is used to get the voltages in the part of the code that reads hits.
The numbers above correspond to the strip values defining the boundaries of the sectors.  Not related to the threshold implementation.  But I moved them to the constants class.

reconstruction/fmt/src/main/java/org/jlab/rec/fmt/hit/Hit.java
zieglerv: But I moved them to the constants class.

baltzell:
is this still relevant?

PR #48, opened 2023-05-23
BMT reading of HV tables on by default.

zieglerv:


raffaelladevita (approved):

baltzell:
Does this require any changes to YAML files?

raffaelladevita:
No, but it will save chefs from setting that variable manually in their yamls, like RG-B is doing for pass2

PR #50, opened 2023-05-26
Use CCDB to determine whether the clock is good, and, if so, use it instead of RCDB's run start time

baltzell:
This is how things need to be done going forward, decoding the correct clock and using it for the Faraday offset calculation.  The gotchas are:

1.  Data that was recorded when the clock's frequency was too large and rolled over within a run.
2.  Data that was decoded without preserving the correct clock.

Currently those are controlled by using the clock frequency from CCDB, currently set for RG-M and later to be usable, and for all previous runs to be unusable, with a threshold of 200 kHz.

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -42,6 +42,7 @@ Nathan Baltzell <baltzell@jlab.org> <baltzell@lappi4.home>
 Nathan Baltzell <baltzell@jlab.org> <baltzell@LappiAir2.home>
 Nathan Baltzell <baltzell@jlab.org> <baltzell@gmx.com>
 Nathan Baltzell <baltzell@jlab.org> <baltzell@llama.jlab.org>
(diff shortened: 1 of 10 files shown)


raffaelladevita (approved):

baltzell:
Difference between gated beam charge using the (old, kludge) raw Faraday cup and the (new, correct) clock for calculating the livetime for the correction for the Faraday cup offset, for a sampling of a few files from ~10 runs from RG-M across the run period and all target/beam combinations, as chosen by Justin, after at least a few minutes into each run.
Screen Shot 2023-05-31 at 11 51 14 AM

baltzell:
And the new version, sampled evenly across ~30% of the files in an entire few-hour run, 15884:
Screen Shot 2023-05-31 at 4 44 45 PM
(The livetime there is the average across the run up to that point, that used for the offset correction.)

PR #56, opened 2023-06-09
Added RG-M AI-assisted tracking NN for run 15058 and 15778 in archive…

raffaelladevita:
… file


raffaelladevita:
To verify the RG-M neural networks were added correctly to the archive, 1000 events from run 15301, file 100 were processed with:
- this branch, selecting the newly added neural networks (run 15058 and 15778)
- the original networks from the RG-M training.

The output files from these two cookings were compared event by event: specifically, the content of REC::Particle and REC::Trajectory were compared event by event finding no differences.

baltzell (approved):

PR #68, opened 2023-07-27
fix: `run-groovy` class path

c-dilks:
Fixes `run-groovy` class paths (`$JYPATH`) for `.jar` file location.


baltzell (commented):
The current "build system" creates a `coatjava` installation directory (within the source tree), with the idea that everything in its `bin` subdirectory should be run only from there.  And then that `coatjava` installation directory can get copied to and used from wherever.  Not that this current approach is good, and it predates me, but this PR would break that.

PR #70, opened 2023-08-01
ci: speedup and parallelize tests

c-dilks:
- Speedup workflow runs by running tests in parallel
- Cancel running workflows upon new commits (on the PR's branch only)


c-dilks:
Not that much of a speedup, possibly because of the additional artifact upload/download time, but it is still good to parallelize the tests to make it easier to add more in the future.

baltzell (approved):

PR #83, opened 2023-08-16
add cli utility for printing json banks

baltzell:


common-tools/clas-io/src/main/java/org/jlab/utils/JsonUtils.java
```diff
@@ -239,5 +262,28 @@ public static DataBank extend(DataEvent event, String bankName, String varName,
         return extend(event, bankName, varName, Map2Json(extension));
     }
 
+    public static void main(String args[]) {
+        OptionParser parser = new OptionParser();
```
raffaelladevita: when printing usage, it shows:

     Usage : undefined -b [bank name]  [input1] [input2] ....
 

I think the "undefined" can be fixed by changing this line to:
OptionParser parser = new OptionParser("hipo-json");

raffaelladevita (changes requested):
Tested successfully displaying the RAW::epics and COAT::config banks

raffaelladevita (approved):

PR #86, opened 2023-08-25
Updated trigger-filter with functionality to use beam current from RAW::epics or a detector raw multiplicity to select events with beam when scalers are unavailable 

raffaelladevita:
Addresses RG-M issue of missing scalers in the first part of the run.

The modifications include:
- trigger bit selection is now based on trigger masks, both to choose the bits that should be active and to veto bits that shouldn't be active;
-  the source of the beam current information can be chosen between the DSC2 scaler readouts from RUN::scaler (default) or a current readout from RAW::epics (selected via the `-s pv-name` command line option);
-  in addition, a threshold on the minimum number of entries in a bank can be applied.

A few notes:
- Epics information is collected from tag-1 events, very similarly to the way scaler information is collected: here the code could be probably unified, creating a generic sequence class but it has not been attempted yet;
- The Epics information is ordered based on Unix time from RUN::config because in some data sets (including RG-M) the RUN::config.timestamp for events with the Epics bank is always equal to -1.
- Comparison of beam current from scalers and Epics suggests that the Epics readout available at a given time applies to an earlier time, with a delay of 1-2 s. For this reason, when choosing to use Epics information, the current threshold is not applied to a single readout but on the minimum within a window of -2 s : +4 s around the closest readout in time.


baltzell (commented):
Let's drop the "Daq" from the new class names.

(Later we should really inherit some of this, shared by DaqScalersSequence, later ...)

baltzell (commented):
FilterBank seems is only about number of rows in a bank?  If so, and it's not planned to become something much more, it would be good to name it less generically, as there's many other things people refer to as "filtering".  For example, FilterBankSize.

baltzell (commented):
I think EpicsSequence.add was copied from a similar implementation supporting Java 8 from before SortedSet/List were fully functional in Java 11?

raffaelladevita:
That’s correct: should I switch?

baltzell (dismissed):

baltzell (approved):

Issue #104, opened 2023-09-18
Workflow runs twice for pull requests


c-dilks:
Commits on a PR branch trigger `maven.yml` twice, e.g., a commit to https://github.com/JeffersonLab/coatjava/pull/103 triggered:

1. https://github.com/JeffersonLab/coatjava/actions/runs/6221158127
1. https://github.com/JeffersonLab/coatjava/actions/runs/6221188205

PR #110, opened 2023-09-19
fix!: use `bash` for `env.sh` to provide access to its true location

c-dilks:
Using `$0` and `$1` in `env.sh` when called as
```bash
. bin/env.sh    # or source bin/env.sh
```
depends on the shell. Ideally, we want:
```
$0 = bin/env.sh
$1 = whatever additional argument was supplied to the '. bin/env.sh' call (see 'run-groovy')
```
However, these may not be the case depending on the shell (especially for `dash`, which may be used in place of `sh`).

Since [there is no POSIX-compliant way to get the true script name](https://stackoverflow.com/questions/29832037/how-to-get-script-directory-in-posix-sh), this PR proposes switching `env.sh` to be in `bash`, so that we may prioritize using `$BASH_SOURCE[0]` in place of `$0` to get the true `bin/env.sh` path.

All `sh` scripts which source `bin/env.sh` are also converted to `bash`.

### DIFF
--- a/bin/bg-merger
+++ b/bin/bg-merger
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 . `dirname $0`/env.sh 
 
--- a/bin/bos2hipo
+++ b/bin/bos2hipo
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 . `dirname $0`/env.sh 
 
--- a/bin/cvt-compare
+++ b/bin/cvt-compare
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 . `dirname $0`/env.sh 
 
--- a/bin/daqEventViewer
+++ b/bin/daqEventViewer
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 . `dirname $0`/env.sh 
 
--- a/bin/dclayereffs-ana
+++ b/bin/dclayereffs-ana
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 . `dirname $0`/env.sh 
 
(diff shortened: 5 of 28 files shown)


c-dilks:
The following scripts may remain in `sh` since they do not source `env.sh`:
```
build-javadocs.sh
bin/dst-maker
validation/advanced-tests/run-eb-tests-all.sh
validation/advanced-tests/run-advanced-tests.sh
validation/advanced-tests/src/eb/scripts/gemc.sh
validation/advanced-tests/src/eb/scripts/gemc-all.sh
```

c-dilks:
Variables in `env.sh` from calling `run-groovy` (which calls `. env.sh groovy`) are:
```
$BASH_SOURCE[0] = coatjava/bin/env.sh
$0              = coatjava/bin/run-groovy
$1              = groovy
$thisEnv        = coatjava/bin/env.sh
$CLAS12DIR      = /home/runner/work/coatjava/coatjava/coatjava
```
Relevant CI log: https://github.com/JeffersonLab/coatjava/actions/runs/6241617294/job/16944369437?pr=109#step:7:10

This is what we expect. 

I've also tested this locally under `bash`, `sh`, and `zsh` and it seems to work correctly.

baltzell (approved):

baltzell:
Seems no github actions ran on this PR?  Resubmit to see if that fixes it.

c-dilks:
Weird. After the rebase, 2 `pull_request` triggers occurred at the exact same time (unlikely related to #104). One [was successful](https://github.com/JeffersonLab/coatjava/actions/runs/6263949021) while the other [cancelled itself](https://github.com/JeffersonLab/coatjava/actions/runs/6263949028) because of the first one. See the checks tab of this PR.

Can you re-open this? (I don't have permission from a fork)

c-dilks:
Rebasing (though CI checks are already re-triggered).

c-dilks:
Successful, and this time only one trigger. The double trigger may have been a GitHub glitch.

Issue #119, opened 2023-09-28
Add `darwin` build testing to CI


c-dilks:
If we want to maintain support for `darwin` (Mac), we should be testing it.

(cf. #118)

PR #120, opened 2023-09-28
fix: avoid using `realpath`

### LINKED ISSUE #118: realpath: command not found on mac
Got this error when running recon-util on a Mac with Monterey version 12.6.9. 

The error originates from ```env.sh```.

The solution was to install with home-brew ```coreutils```, as suggested by many posts on the web. The same posts report that ```realpath``` may or may not be available on Mac depending on the version of the operating system and that it is available by default (only) starting from MacOS 13 (Ventura).

Is there an alternative command that is less system-dependent?

c-dilks:
close #118 

This uses `cd` and `pwd -P` to get the full absolute path, rather than `realpath` which is not available on older MacOS versions.

### DIFF
--- a/bin/env.sh
+++ b/bin/env.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 thisEnv=${BASH_SOURCE[0]:-$0}
-export CLAS12DIR=$(realpath $(dirname $thisEnv)/..)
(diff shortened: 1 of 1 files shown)


c-dilks:
@raffaelladevita, could you test if this fix resolves #118?

Meanwhile I will automate such testing for #119.

baltzell (approved):

raffaelladevita:
I confirm it works for me!

PR #127, opened 2023-10-09
ci: add final job, to be required as a trunk branch protection rule

c-dilks:
Github branch protection rules for a job matrix status check require _every_ matrix element to be set to "required". This PR adds a simple downstream job, named `final`, which will only run _successfully_ if all tests pass.

@baltzell: with this PR, the `development` branch protection rule may be set to require just the `final` status check to pass


c-dilks:
After some testing, I find that a skipped/cancelled job is not really considered a failure, according to branch protection rules, so the conditionals here in `final` force it to always run and only succeed or fail; cancellation will cause its failure.

baltzell (approved):

PR #131, opened 2023-10-09
style: use block scalar for validation dispatch payload

c-dilks:
Fold and chomp the `json` payload with block scalar syntax `>-`, instead of confusing some editors or parsers with a multi-line single quote.


baltzell (dismissed):

c-dilks:
rebased and resolved conflict from #130

baltzell (dismissed):

baltzell (approved):

PR #138, opened 2023-10-13
fix: payload branch and URL for `push` dispatch triggers

c-dilks:
- fixes the branch name for `push` triggers, which was not correct and caused `git clone` failures in `clas12-validation` runs
- adds a URL to the relevant commit for `push` triggers, which is usually a PR merge


c-dilks:
This one should finally fix the dispatched workflows for PR merges.... except now today `gemc` is not building (unrelated issue and has been communicated).

baltzell (approved):

PR #141, opened 2023-10-16
style: run `dos2unix` to remove Windows newlines

c-dilks:
Remove the extra carriage return, commonly inserted by Windows text editors. They can often cause issues in parsed files, so this PR just removes them all.

When viewing the `diff` of this PR, be sure to check "Hide Whitespace" or it looks like I changed way more than I did!

### Procedure:
```bash
rg -l '\r' | xargs -I{} dos2unix {}
```
where `rg` is [`ripgrep`](https://github.com/BurntSushi/ripgrep).


baltzell (approved):
Found the "Hide whitespace" under a gear icon, 
Screen Shot 2023-10-17 at 6 05 27 PM

PR #146, opened 2023-10-18
ci: replace `clas12-validation` dispatch with reusable workflow

c-dilks:
Since other repositories (`clas12Tags`, `clas12-config`) would benefit from dispatching `clas12-validation` runs, this PR changes the dispatch workflow to a reusable workflow caller.

Depends on https://github.com/JeffersonLab/clas12-validation/pull/33

This PR resolves _several_ issues:
- PRs from forks of `coatjava` now successfully trigger `clas12-validation` workflow runs
- no more [annoying queuing issues](https://github.com/JeffersonLab/clas12-validation/issues/10): rapid commits on a PR branch will successfully cancel running `clas12-validation` workflow runs
- no need for an access token secret
- it is easier to find the running `clas12-validation` workflow run, since we are no longer dispatching; the reusable workflow means the workflow run happens _here_ and is fully controllable _here_
- the status badge in `README.md` is now accurate: it reflects the status of `coatjava` calls of `clas12-validation` workflow, instead of the _overall_ status of `clas12-validation` workflow runs (which may likely be independent of `coatjava`'s status)
- significantly simpler caller workflow `validation.yml`: very unlikely the important part of this file will need to change, so this file may simply be added to other repos (`clas12Tags` and `clas12-config`) so they may also run `clas12-validation`

### DIFF
--- a/.github/workflows/validation.yml
+++ b/.github/workflows/validation.yml
@@ -11,8 +11,13 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-
   clas12-validation:
     name: clas12-validation workflow
-    uses: JeffersonLab/clas12-validation/.github/workflows/validation.yml@reusable-workflow  # FIXME: use `@main`
-    secrets: inherit  # FIXME: may not be needed if we can avoid `trigger-workflow-and-wait`
+    uses: JeffersonLab/clas12-validation/.github/workflows/ci.yml@reusable-workflow  # FIXME: use `@main`
+    with:
+      # num_events: 0 # FIXME test this
+      config_file_versions: >-
+        {
+          "coatjava": "latest",
+          "gemc":     "latest"
+        }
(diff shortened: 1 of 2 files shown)


c-dilks:
All right, this seems to be working now, according to the [last workflow run](https://github.com/JeffersonLab/coatjava/actions/runs/6582255485) and various other tests.

I'm now going to remove the `clas12-validation` feature branch specification, so:
- the next workflow run will fail, since the reusable workflow is not on `clas12-validation@main` yet
- retrigger the workflow when https://github.com/JeffersonLab/clas12-validation/pull/33 is merged, then merge this PR as soon as it is successful

c-dilks:
> the next workflow run will fail, since the reusable workflow is not on `clas12-validation@main` yet

Validation workflow didn't even run, so it looks like everything passed :grin:

baltzell (approved):

PR #157, opened 2023-11-29
fixed the issue with the file name when space is introduced in the YA…

gavalian:
…ML file. Now it will trim the filename before loading the network


raffaelladevita (commented):
What happens if the path points to a file that doesn't exist?

gavalian:
Most likely fails to load the file. What is the preferred reaction?  Halt?

raffaelladevita:
Yes, it would be good to halt

baltzell (approved):

PR #167, opened 2024-01-17
EVIO event ordering for helicity sequence detection

### LINKED ISSUE #166: Helicity sequence detection
Looks like EVIO event misordering is significantly larger in RG-K 2024 than previously, enough that the current helicity sequence detection method is too inefficient.  Ultimately it will be superceded by the new helicity decoder board, but it's worthwhile to have sorting functionality at least until that's vetted.

baltzell:


baltzell:
Things to consider ...
* possibility of duplicated event numbers from CODA (currently only one would be kept)
* preserve events with no HEAD bank or negative event numbers (currently dropped)
* preserve the original event order in HIPO and only use ordering for sequence detection and generating the HEL::flip bank (currently HIPO output is reordered)

Meanwhile, as is, this does appear to address the issue with recent data:
Screen Shot 2024-01-17 at 1 26 21 PM

PR #174, opened 2024-01-22
Issue 173 helicity decoder

mholtrop:
This code allows semi-correct (*) decoding for the "old" style data, were 40 helicity decoder banks show up in one event, and the correct type data where this bank is in each event.
(*) Semi-correct because it will only parse one of those 40 helicity decoder entries.

The code was tested on run 19400 file 40, which has helicity decoder data every event, and 19407 which has 40 in one event and no crash or error is observer. 

Please double check the parsing of the timestamp information.

### DIFF
--- a/common-tools/clas-detector/src/main/java/org/jlab/detector/decode/CodaEventDecoder.java
+++ b/common-tools/clas-detector/src/main/java/org/jlab/detector/decode/CodaEventDecoder.java
@@ -1049,34 +1049,53 @@ public class CodaEventDecoder {
                     
                     long[] longData = ByteDataTransformer.toLongArray(node.getStructureBuffer(true));
                     int[]  intData  = ByteDataTransformer.toIntArray(node.getStructureBuffer(true));
-                    long  timeStamp = longData[2]&0x0000ffffffffffffL;
-                    
-                    int tsettle  = DataUtils.getInteger(intData[16], 0, 0) > 0 ? 1 : -1;
-                    int pattern  = DataUtils.getInteger(intData[16], 1, 1) > 0 ? 1 : -1;
-                    int pair     = DataUtils.getInteger(intData[16], 2, 2) > 0 ? 1 : -1;
-                    int helicity = DataUtils.getInteger(intData[16], 3, 3) > 0 ? 1 : -1;                             
-                    int start    = DataUtils.getInteger(intData[16], 4, 4) > 0 ? 1 : -1;                             
-                    int polarity = DataUtils.getInteger(intData[16], 5, 5) > 0 ? 1 : -1;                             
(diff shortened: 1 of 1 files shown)


baltzell:
Looks like I accidentally posted this on the corresponding issue instead of this PR ...

Running this on one of the few files I'd been using for testing (run 19349, file 40), looks to me like the bit sequences are now off in the HEL::decoder bank.  For example:

```
Pair:      11011101110111011101110111011101
Pattern:   00000000000000000000000000000110
Helicity:  01010101010101010101010101010101
HelicityP: 10011010010110011001101001011010
```

Previously they were all consitent with good quartet patterns, assuming signal inversion (which has now been confirmed).

common-tools/clas-detector/src/main/java/org/jlab/detector/decode/CodaEventDecoder.java
```diff
@@ -1049,34 +1049,60 @@ public HelicityDecoderData getDataEntries_HelicityDecoder(EvioDataEvent event){
...
+                    if(((int) (( ((long)intData[i_data_offset]) & 0x00000000ffffffffL ) >> 27)) != 0x18){
+                        System.err.println("ERROR:  HelicityDecoder data is corrupted. DECODER BANK not found.");
+                        return null;
+                    }
+                    int num_data_words = intData[i_data_offset]&0x07ffffff;
+                    if(num_data_words < 14){
+                        System.err.println("ERROR:  HelicityDecoder data is corrupted. Not enough data words.");
+                        return null;
+                    }
+                    int tsettle  = DataUtils.getInteger(intData[i_data_offset+9], 0, 0) > 0 ? 1 : -1;
```
raffaelladevita, line 1081: i_data_offset should be incremented by 1 before this line

raffaelladevita (changes requested):

common-tools/clas-detector/src/main/java/org/jlab/detector/decode/CodaEventDecoder.java
mholtrop: Thank you for catching that Raffaella. You are quite correct. An update was pushed.
I still need to do a detailed analysis of the helicity. I am not quite understanding what Nathan is referring to.

baltzell: I'm referring to the most critical data in the HEL::decoder bank, those "Array" variables that contain the helicity state sequences seen so far.  Previously they were consistent with good sequences, and, in this PR prior to Raffaella's recent suggestion, were not.

baltzell:
The resulting HEL::decoder bank looks good again regarding helicity bit sequences.

raffaelladevita (approved):

PR #175, opened 2024-01-25
ci: bump action versions

c-dilks:
Fixes Node.js deprecation warnings (which are starting to appear in other repos).


baltzell (approved):

c-dilks:
Another sporadic coatjava build failure. This is the 3rd one today...

PR #178, opened 2024-01-29
hotfix: trigger validation with GEMC tag 5.4

c-dilks:
Temporarily keep GEMC at 5.4; will revert to 'latest tag' when 5.6 is working.


c-dilks:
Actually... nevermind. I'm going to fix validation at 5.4 for all from within `clas12-validation` instead.

PR #182, opened 2024-02-06
Changed ALERT bank names back to original and added option to...

mpaolone:
...read from AHDC::adc to populate hit bank


mpaolone:
missing comma in data.json: fixing now

PR #193, opened 2024-02-16
Fix bank names in schema generation

baltzell:
and add error checking in schema generator script


raffaelladevita (changes requested):
TimeBasedTrkg::AITrajectory can be removed from dcalign

raffaelladevita (approved):

PR #199, opened 2024-02-22
Fix data bank name in AHDCengine

efuchey:
Fixed the data bank name in AHDCengine from AHDC::tdc to AHDC::adc (which is used by the corresponding hit reader)


c-dilks:
Validation checks are failing for unrelated reasons (problem with the GEMC container), so you may ignore them for this PR.

baltzell (approved):

PR #203, opened 2024-02-29
Updated target geometry and material handling in CVT Eloss 

### LINKED ISSUE #185: add polarized target material budget for eloss in CVT tracking

raffaelladevita:
Needed for RG-C

* Target geometry and materials are now read from the CCDB table /geometry/materials/target. The target volumes are defined in the table according to the following rules:
  * surfaces are identified by layer and should be defined for layer = 1, 2, and 3,
  * multiple sublayers, identified by component, can be added with different materials,
  * layer=1 component=1 is the target material and should always be present,
  * the target material is assumed to be a full cylinder  (min radius should be 0) while the other surfaces are "tubes",
  * for layer 1, the target material volume is created and any additional sublayer is assumed to be equivalent to the target cell walls,
  * for layer 2 and 3, when multiple sublayers are defined, the surface radius and length are set to the largest values among the sublayers,
* ELoss in the target cell accounts for the vertex position and, therefore, the raster in the RG-C case.
* Geometry class singleton now loads geometry for the actual run number.


raffaelladevita:
Tested by Noemie on RG-A and RG-C geometries
[updated_geom_RGA.pdf](https://github.com/JeffersonLab/coatjava/files/14449529/updated_geom_RGA.pdf)
[updated_geom_RGC.pdf](https://github.com/JeffersonLab/coatjava/files/14449530/updated_geom_RGC.pdf)

baltzell (approved):

PR #216, opened 2024-03-07
Fix old EB test

baltzell:
Don't know when this broke, nor why it wasn't showing up in the CI builds.  Probably only shows up when running maven from the reconstruction directory.


baltzell:
I just copied what the DC one was doing to "fix" it.

raffaelladevita (approved):

PR #224, opened 2024-03-18
modified the HipoToHipoWriter to accept wildcard for banks in the YAM…

gavalian:
Now one can write a wildcard in the YAML file to filter out the banks, there no need for creating schema directory


baltzell:
What's the new engine for?

gavalian:
Oh, that's where my engines were, these are to export clusters from hit-based training to be used for AI training and validation.

baltzell:
need to regenerate this with only the wanted changes

PR #243, opened 2024-04-02
ci: download validation data from XRootD server

c-dilks:


gavalian (dismissed):

c-dilks:
@baltzell Ubuntu's `xrootd-client` seems to work out of the box, no need to set `LD_PRELOAD`, so it should be straightforward to make this work with our CI.

c-dilks:
Closing this, unless we have a need for this here. Note that we are now using such validation files in:
- https://github.com/JeffersonLab/iguana/pull/188
- https://github.com/JeffersonLab/clas12-timeline/pull/204

It's not (yet) clear what we can use the `.h5` files for.

Issue #263, opened 2024-04-23
Add DC::jitter and RECHB banks to dchv schema


hauenst:
I realized that the new dchv schema is missing the RECHB banks and DC::jitter for purpose of using that schema for special DC studies.

PR #264, opened 2024-04-23
Add DC::jitter to dchv schema. Add RECHB banks to dchv by replacing d…

hauenst:
…st with dsthb in list

Add the missing banks to DC HV schema (see open issue #263 )


c-dilks:
Don't worry about the CI failure, it's a new bug. I'll fix it in another PR.

baltzell (approved):

PR #289, opened 2024-07-07
remove unused dependency

baltzell:


c-dilks:
`org.jlab.coda.xmsg` is used in `common-tools/clas-io`, is it found some other way?

baltzell:
> `org.jlab.coda.xmsg` is used in `common-tools/clas-io`, is it found some other way?

looks like it's just included in a pom.xml and not really used

c-dilks (approved):

Issue #293, opened 2024-07-11
freehep maven dependency issues


baltzell:
Started probably yesterday ...

[ERROR] Failed to execute goal on project clas12rec: Could not collect dependencies for project org.jlab.clas:clas12rec:pom:10.1.2-SNAPSHOT
[ERROR] Failed to read artifact descriptor for org.freehep:freehep-jminuit:jar:1.0.2
[ERROR] 	Caused by: The following artifacts could not be resolved: org.freehep:freehep-jminuit:pom:1.0.2 (absent): Could not transfer artifact org.freehep:freehep-jminuit:pom:1.0.2 from/to freehep-repo-public (https://srs.slac.stanford.edu/nexus/content/groups/freehep-maven2-public/): Connection reset

gavalian:
I'm getting the same thing. I wonder if this is a temporary problem? 
If not, we can install the Jminuit locally on clas12maven and use it from there.

c-dilks:
I can't access https://srs.slac.stanford.edu/nexus/content/groups/freehep-maven2-public/ and all the recent wayback machine snapshots get stuck in a loop of errors. Maybe we use a different freehep maven repo mirror?

gavalian:
If you can find one, yes. all the documentation on this is outdated. I'm having hard time finding the right one.

c-dilks:
I tried using the one from mvnrepository: https://mvnrepository.com/artifact/org.freehep/freehep-jminuit

I also removed everything that has to do with srs.slac.stanford.edu, but _something_ is still dependent on it... I'll open a draft PR so we can take a closer look at the logs.

c-dilks:
We decided to wait til Monday to see if `srs.slac.stanford` comes back; until then, nightly `clas12-validation` failures will spam our inboxes...

baltzell:
I installed all the necessary freehep maven repos on clasweb ...

PR #295, opened 2024-07-11
fix(build): use `mvnrepository` to get `freehep-jminuit`

c-dilks:
## do not merge!

Trying to debug #293


c-dilks:
close in favor of #296

PR #303, opened 2024-07-29
fix!: `env.sh` sourceable should not be in `$PATH`

c-dilks:
Having `env.sh` in one's `$PATH` means that when some shells run `source env.sh` from anywhere, it may prioritize the `env.sh` from the `coatjava` installation, rather than the `env.sh` in the caller's present working directory. Since both `coatjava`'s `env.sh` and the caller's `./env.sh` may be silent, this can lead to confusion as to why the local `./env.sh` is not working.

This PR moves `env.sh` to `libexec/`, which should not be in one's `$PATH`. Any downstream code which calls `source env.sh` may be affected, which is why the conventional commit label is `fix!` rather than `fix` (_i.e._, potential breaking change).


validation/advanced-tests/run-advanced-tests.sh
```diff
@@ -31,6 +31,7 @@ esac
 chmod +x install-claracre-clas.sh
 echo Y | ./install-claracre-clas.sh -f 5.0.2 -j 11 -l local
 if [ $? != 0 ] ; then echo "clara installation error" ; exit 1 ; fi
+cp -rp ../../coatjava/libexec $CLARA_HOME/plugins/clas12/ # FIXME: should go in `install-claracre-clas.sh`, or can we use `../../install-clara` instead?
```
c-dilks: This is a kluge; we need a better approach here

c-dilks: perhaps we kluge now and remove this test in a separate PR.

c-dilks: as discussed, let's use `../../install-clara`

baltzell:
Note, this means the old CLARA installer will not result in a working COATJAVA installation.  But that's acceptable since it also doesn't provide a working CLARA anymore.

baltzell (approved):

baltzell (approved):

PR #308, opened 2024-08-13
java 17 and javafx 22

baltzell:
supersedes #300 and #301

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -21,7 +21,7 @@ defaults:
     shell: bash
 
 env:
-  java_version: 11
+  java_version: 17
   java_distribution: zulu
   groovy_version: 4.0.3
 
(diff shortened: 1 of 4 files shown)


gavalian:
I'm now working to incorporate InstaRec online, and want to experiment with virtual threads, available only in JDK 21.  Can we switch to 21 instead? If not, I will come up wth workaround.

c-dilks:
> I'm now working to incorporate InstaRec online, and want to experiment with virtual threads, available only in JDK 21.  Can we switch to 21 instead? If not, I will come up wth workaround.

21 also has the new foreign function and memory API, which may be useful for Iguana bindings. JNI should be sufficient though, if we want to stay with 17.

gavalian:
My immediate needs are threads, but I had my eye on foreign functions too. There is no point in using JNI if Java is deprecating them. Also, 21 is the current version, how long should we wait until transitioning to it if not now?

gavalian (approved):

PR #310, opened 2024-08-16
Add background merging engine

baltzell:


reconstruction/bg/src/main/java/org/jlab/service/bg/BackgroundEngine.java
```diff
@@ -0,0 +1,83 @@
...
+        Boolean suppressDoubles = Boolean.valueOf(getEngineConfigString(CONF_SUPPRESS_DOUBLES,"true"));
+        Boolean preserveOrder = Boolean.valueOf(getEngineConfigString(CONF_PRESERVE_ORDER,"true"));
+        bgmerger = new EventMerger(detectors.split(","), orders.split(","), suppressDoubles, preserveOrder);
+        openNextFile();
+        return true;
+    }
+
+    private void openNextFile() {
+        String filename = bgfilenames.remove();
+        bgfilenames.add(filename);
```
raffaelladevita, line 63: it would be nice to have an option to turn it off, so to reproduce the behavior of the standalone bg-merger

raffaelladevita (approved):

PR #312, opened 2024-08-20
fix issue for path length in FD tracking

tongtongcao:
The original point is not starting from vertex when set path length for hits on DCs. It causes that time of flight is wrong, and further causes that hit time is wrong, and finally causes that doca is wrong.
After the issue is fixed, relative parameters in CCDB need to be re-calibrated.

### DIFF
--- a/common-tools/clas-tracking/src/main/java/org/jlab/clas/tracking/kalmanfilter/zReference/KFitter.java
+++ b/common-tools/clas-tracking/src/main/java/org/jlab/clas/tracking/kalmanfilter/zReference/KFitter.java
@@ -708,7 +708,9 @@ public class KFitter extends AKFitter {
             return false;
         }
     }
-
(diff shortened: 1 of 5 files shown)


zieglerv (dismissed):

tongtongcao:
Path length, as a conventional parameter for state vector in tracking, should be kept. Since no vertex information directly from tracking, starting point for path length in tracking is the final point at the last measurement DC layer. After vertex is obtained later, transition for the starting point from the final point to vertex is taken, so that time of flight can be calculated by path length from vertex to each DC hit. Comments for the above explanation are added into codes.

baltzell:
can someone find which merged PR included this?

tongtongcao:
This PR has been merged into PR https://github.com/JeffersonLab/coatjava/pull/494

PR #314, opened 2024-08-21
imported AI-assisted tracking network for RG-C Fall2022: network trai…

raffaelladevita:
…ned on run 17277 and assigned to run range 16843-17408


raffaelladevita:
The updated list of run ranges in the network file is as follows:
Screenshot 2024-08-21 at 2 31 22 PM
The modifications are for the run range 16843-17408. 

A test was done processing 100 events with this branch, loading the network from the updated file, and with 10.1.0, loading the network from /work/clas12/jnp/rgb/clas12rgc.network, finding no differences

baltzell (approved):

PR #322, opened 2024-08-26
enable rebuilding scalers in post-processing engine

baltzell:


etc/services/data-ai-pp.yaml
raffaelladevita, line 1: maybe change the name to data-pp.yaml since there is no ai

raffaelladevita (changes requested):

raffaelladevita (approved):

PR #324, opened 2024-08-28
fix: rename `hipo-add` to `hipo-merge-histograms` and clarify usage

### LINKED ISSUE #323: `hipo-add` should give a clear error message when used on data files
On data files (not HIPO files filled with histograms), running:
```bash
hipo-add -o out.hipo in1.hipo in2.hipo
```
Outputs _many_ lines repeating
```
(warning) :>>> no node present with group = 21, item = 7 positoin = -1
[hipo node] --> error : (1,1) requested type=INT has type=FLOAT
[hipo node] --> error : (1,1) requested type=INT has type=FLOAT
[hipo node] --> error : (1,1) requested type=INT has type=FLOAT
[hipo node] --> error : (1,1) requested type=INT has type=FLOAT
```
and eventually an empty HIPO file.

This is actually a `groot` problem, since `hipo-add` wraps `org.jlab.groot.data.TDirectory`. Note that for data files, run instead:
```
hipo-utils -merge -o out.hipo in1.hipo in2.hipo
```
It would be helpful for those of us who forget the difference between `hipo-add` and `hipo-utils -merge` to have better error reporting or usage guidance.

c-dilks:
Resolves #323

### DIFF
--- a/bin/hipo-add
+++ b/bin/hipo-add
@@ -1,8 +1,0 @@
-#!/bin/bash
-
-. `dirname $0`/../libexec/env.sh
-
-java -Xms1024m \
-    -cp "$CLAS12DIR/lib/clas/*:$CLAS12DIR/lib/plugins/*" \
-    org.jlab.groot.data.TDirectory \
-    $*
--- a/bin/hipo-merge-histograms
+++ b/bin/hipo-merge-histograms
@@ -1,0 +1,15 @@
+#!/usr/bin/env bash
+
+. `dirname $0`/../libexec/env.sh
+cmd="java -Xms1024m -cp $CLAS12DIR/lib/clas/*:$CLAS12DIR/lib/plugins/* org.jlab.groot.data.TDirectory"
+
+if [ $# -eq 0 ]; then
+  echo """
+  hipo-merge-histograms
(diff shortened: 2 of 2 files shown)


baltzell:
maybe best to just remove it altogether?

c-dilks:
> maybe best to just remove it altogether?

That was my original idea, but `hipo-utils -merge` does not actually add histograms (even though it returns exit code 0). On the other hand, `hipo-add` / `hipo-merge-histograms` does. So we have:

- `hipo-utils -merge` for combining _data_ files (e.g., DST files)
- `hipo-merge-histograms` for combining files with _histograms_

On the other hand, ROOT's `hadd` does both of these, which is why this behavior confused me.

A better solution would be to have `hipo-utils -merge` add the histograms correctly, but that's an upstream issue (`org.jlab.jnp.hipo4.utils.HipoUtilities`, cc @gavalian)

gavalian:
Merge is a function of HIPO library, which just merges events. However, the hadd is a function of GROOT which has to read and interpret histogram classes and add the content. The HIPO library does not have histogram implementations and can not read and parse the content. That's why this functionality is kept separate.

baltzell (approved):

PR #325, opened 2024-08-29
fix: allow `run-groovy` users to set `JAVA_OPTS`

c-dilks:
`libexec/env.sh` hard codes a value of `$JAVA_OPTS` for groovy calls, (via `run-groovy`) making it impossible for `run-groovy` users to change any Java options, e.g., heap size limits. This PR allows users to override `env.sh`'s settings by appending user-level `JAVA_OPTS`.

### DIFF
--- a/libexec/env.sh
+++ b/libexec/env.sh
@@ -30,7 +30,7 @@ if [ $# -ge 1 ]; then
     # additional variables and settings for groovy
     if [ "$1" = "groovy" ]; then
       JYPATH="${JYPATH:+${JYPATH}:}${CLAS12DIR}/lib/packages"
-      export JAVA_OPTS="-Dsun.java2d.pmoffscreen=false -Djava.util.logging.config.file=$CLAS12DIR/etc/logging/debug.properties -Xms1024m -Xmx2048m -XX:+UseSerialGC"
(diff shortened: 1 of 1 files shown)


c-dilks:
So for example, now this will work
```bash
JAVA_OPTS="-Xms256m -Xmx256m" run-groovy .....
```

baltzell:
We should reassess what this "run-groovy" script is supposed to do.

c-dilks:
> We should reassess what this "run-groovy" script is supposed to do.

- `JYPATH=.....` => this is the classpath to include coatjava, and is the primary reason we have `run-groovy` (_cf_. `run-jshell`)
- `-Dsun.java2d.pmoffscreen=false` => not sure why we use this; seen also in `dclayereffs-ana`
- `-Djava.util.logging.config.file=$CLAS12DIR/etc/logging/debug.properties` => somewhat common in wrapper scripts, seems useful
- `-Xms1024m -Xmx2048m` => in principle we could remove these and trust the defaults, but this PR allows the caller to override them; probably better to keep these here so users aren't surprised by 'out of memory' errors on slurm
- `-XX:+UseSerialGC` => looks like all our `java` wrappers in `bin/` use this (default is parallel GC)

baltzell (approved):

PR #328, opened 2024-09-05
create DC super layer GEANT4 volumes

### LINKED ISSUE #259: fix DC geometry for simulation to make cells hexagonal

raffaelladevita:
needed to make DC cells in gemc hexagonal


raffaelladevita:
See https://clasweb.jlab.org/wiki/images/b/bd/20240516_dcgeometry.pdf

baltzell (approved):

common-tools/clas-jcsg/src/main/java/org/jlab/detector/geant4/v2/DCGeant4Factory.java
```diff
@@ -581,7 +581,7 @@ public final class DCGeant4Factory extends Geant4Factory {
     private int nsgwires;
 
     private final double y_enlargement = 3.65;
-    private final double z_enlargement = -2.96;
+    private final double z_enlargement = -2.46;
```
raffaelladevita, line 584: @hauenst: This parameter is only used to define the region volumes that contain the superlayer volumes. The thickness of the region volumes had to be increased after the superlayer thickness was fixed, accounting for the 25 deg tilt in the trapezoid definition. The region volumes are virtual, so this change is not really modifying the physical volumes but just avoiding an overlap issue.

PR #346, opened 2024-10-18
Use real run numbers for geometry

### LINKED ISSUE #59: Need/want a geometry manager
To facilitate moving to using run number instead of variation for geometry, which means moving access from ReconEngine.init() to processEvent, which means caching and thread-safety and etc, hence the desire for a "manager".

baltzell:
Added abstract method detectorChanged(int run) to ReconstructionEngine, which is called automatically when the run number changes.

Moved every engine's geometry initialization from init or elsewhere to detectorChanged.

Changed the name of ReconstructionEngine's abstract event-processing method from processDataEvent to procesDataEventUser, and filterEvent to processDataEvent, such that all external uses of processDataEvent still do the right thing.

And removed EC's run number check per Cole.

Closes #59


baltzell:
Test failing on EC photon efficiency for run 11.  Cole says it's due to remaining CCDB changes needed (triggered by the removal of "isMC" run number check), which was anticipated.  Retrigger CI after fixing CCDB.

baltzell:
test files regenerated with latest gemc/dev, still no reconstructed photons ...

baltzell:
From the CI logs for the electrongamma test:
Screenshot 2025-01-03 at 16 23 52

So, 88% of photons are getting misidentified as neutrons.  (The test threshold is so far 84% photon efficiency).

baltzell:
And the electronneutron test:
Screenshot 2025-01-03 at 16 27 07

This is more of a distraction, due to different requirements in the efficiency calculations for the matrix and the actual test failure, and upon closer look probably has the same cause as the photon failure (timing).

baltzell:
github appears broken, closing to reopen

PR #348, opened 2024-10-23
Use clas12-config for CI gcards

baltzell:


c-dilks:
I'm probably missing something here, but I don't see `gemc.sh` being called in the CI. `gemc-all.sh` calls it, and as far as I can tell, nothing calls `gemc-all.sh`.

validation/advanced-tests/src/eb/scripts/gemc.sh
```diff
@@ -11,9 +11,12 @@ then
     exit
 fi
 
+v=5.10
+git clone https://github.com/jeffersonlab/clas12-config
```
c-dilks: FYI, you can get the latest version of any `gcard` with
```
clas12-config/util/latest.rb clas12-config/gemc clas12-default
```
which returns
```
clas12-config/gemc/5.10/clas12-default.gcard
```
Run `latest.rb` with no args for guidance. In this way, you don't have to hard code a version number here.

c-dilks (dismissed):

validation/advanced-tests/src/eb/scripts/gemc.sh
baltzell: Yes, in part for speed, GEANT/GEMC isn't run in this repo's CI jobs themselves.  But the HIPO files it generates are used as inputs.

c-dilks (approved):

PR #351, opened 2024-10-24
Iss153 cvt straight trk angle

zieglerv:
Fix for straight tracks validated: https://clasweb.jlab.org/wiki/images/7/76/20241023-cvtstraighttrack.pdf


reconstruction/cvt/src/main/java/org/jlab/rec/cvt/measurement/Measurements.java
```diff
@@ -297,7 +297,15 @@ private List<Surface> getClusterSurfaces(DetectorType type, List<Cluster> cluste
             if(cluster.getDetector()!=type) continue;
             int layer = MLayer.getType(type, cluster.getLayer()).getCVTLayer();
             Surface measure = cluster.measurement();
-            measure.hemisphere = Math.signum(cluster.center().y());
+            if(cluster.getDetector()==DetectorType.BMT) {
```
raffaelladevita: isn't this necessary only for BMTC?

raffaelladevita: it may be better to clone this method adding seed as an argument like it is done for the cosmic ray

reconstruction/cvt/src/main/java/org/jlab/rec/cvt/services/TracksFromTargetRec.java
```diff
@@ -191,15 +191,19 @@ public List<Track> getTracks(DataEvent event, boolean initFromMc, boolean kfFilt
         kf2.polarity = (int) Math.signum(Constants.getSolenoidScale());
         Measurements measure = new Measurements(xb, yb, Constants.getInstance().kfBeamSpotConstraint());
         for (Seed seed : this.CVTseeds) { 
+            measure.setSeed(seed);
```
raffaelladevita: this should not be necessary because seed is the argument of getMeasurement at line 199

raffaelladevita (changes requested):

reconstruction/cvt/src/main/java/org/jlab/rec/cvt/measurement/Measurements.java
zieglerv: yes. And done...

reconstruction/cvt/src/main/java/org/jlab/rec/cvt/services/TracksFromTargetRec.java
zieglerv: no longer necessary after changes implemented in comment above.  So it is removed.

reconstruction/cvt/src/main/java/org/jlab/rec/cvt/measurement/Measurements.java
zieglerv: Changes done.

raffaelladevita (approved):

PR #357, opened 2024-10-28
doc: banks' group and item IDs

c-dilks:
- add a README with all the bank IDs
- add the script which generated this README from the JSON files


c-dilks:
We could automate generating this README file, or even deploy it to Pages, but these JSON files change so rarely such that it isn't worth the effort.

baltzell (dismissed):
But I think the clas6 banks should be removed from coatjava

c-dilks:
Updated w.r.t. #359.

baltzell (approved):

PR #374, opened 2024-11-13
Update RICH support structure STL files

cpecar:
This PR will update the STL files representing the aluminum, CFRP, and mirror support elements for the RICH. The STL files were extracted from an updated simplified CAD model from Frascati engineers.


cpecar:
Currently investigating strange failures of an overlap check in gemc, stating that there are overlaps on the order of 10cm between volumes that should have significant separation. Will investigate further before marking ready for review.

cpecar:
Overlaps checked with Mauri, now suspected to instead be an issue with how geant4 is handling boolean operations used for the spherical mirror volumes (which are native geant4, see https://github.com/gemc/clas12Tags/issues/51). I have now confirmed that these new STL volumes and the overlap reports are not causing any issues when actually running gemc, so this should be ready to merge.

baltzell:
Should this have zero effect on RICH reconstruction?  If so, maybe Marco would like to confirm it?

cpecar:
I do not think this should have any impact but you are right that we should confirm this with @mcontalb. I will also try to test this today.

baltzell:
Let's go ahead and merge it, we can revert if necessary.

c-dilks (approved):

PR #377, opened 2024-11-19
feat: merge a set of HIPO files to a smaller set

c-dilks:
This tool merges a set of input HIPO files to a set of output HIPO files, where you may control the number of input files per output file; for example, use this tool if you have 1000 small HIPO files but would rather have 10 large HIPO files.

### DIFF
--- a/bin/hipo-multi-merge
+++ b/bin/hipo-multi-merge
@@ -1,0 +1,107 @@
+#!/usr/bin/env ruby
+
+require 'optparse'
+require 'ostruct'
+require 'fileutils'
+
+def print_log(name, val)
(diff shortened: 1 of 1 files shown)


gavalian:
Hipo utils can split the file into smaller chunks, so one could merge a file and then split it into smaller chunks.

c-dilks:
> Hipo utils can split the file into smaller chunks, so one could merge a file and then split it into smaller chunks.

Yes, but this PR's script is more efficient since it just needs to run the merge commands; it can do so either sequentially locally or in parallel with Slurm.

baltzell (approved):

Issue #373, opened 2024-11-06
Restore the possibility to merge background multiple times on the same file


raffaelladevita:

raffaelladevita:
addressed by #390

PR #390, opened 2024-12-12
fix issues with running bg-merger twice and add more functionalities

raffaelladevita:
* address #373 
* new functionalities:
  * accept a list of bg files
  * loop over the same bg file
  * merge multiple events to increase bg level
  * save bg-removed bg events
  * allow the selection of specific DC regions for merging


baltzell:
Did we lose the error return value in the engine's init?

baltzell:
Let's make sure the default settings, e.g., which detectors are enabled, are what most people should want (and what should be used on OSG).

raffaelladevita:
> Let's make sure the default settings, e.g., which detectors are enabled, are what most people should want (and what should be used on OSG).

done

raffaelladevita:
> Did we lose the error return value in the engine's init?

The error is now returned by the setgFiles method

baltzell:
Is engine's init always returning good now?

raffaelladevita:
No, it’s returning false if any of the input files doesn’t exist or is unreadable or…as before

baltzell (approved):

PR #419, opened 2025-01-08
DC clustering updates

tongtongcao:
Main updates include:
1. Fix a bug in the splitter for complicated hit clumps, and update resolver for hit-overlapped clusters from the splitter.
2. Loose limit for clusters with consideration of real hit lost due to dead strips, AI-denoising, edge effect, etc. 
3. Fix issues for hits shared by multiple clusters.
See details from 
[New DC Clustering.pdf](https://github.com/user-attachments/files/18349976/New.DC.Clustering.pdf)

### DIFF
--- a/common-tools/coat-lib/pom.xml
+++ b/common-tools/coat-lib/pom.xml
@@ -64,13 +64,13 @@
     <dependency>
         <groupId>j4ml</groupId>
         <artifactId>j4ml-neuroph</artifactId>
-        <version>1.0</version>
+        <version>0.9-SNAPSHOT</version>
     </dependency>
 
     <dependency>
         <groupId>j4ml</groupId>
         <artifactId>j4ml-clas12</artifactId>
-        <version>1.0</version>
+        <version>0.9-SNAPSHOT</version>
     </dependency>
 	
     <dependency>
--- a/common-tools/coat-lib/pom.xml
+++ b/common-tools/coat-lib/pom.xml
@@ -64,13 +64,13 @@
     <dependency>
         <groupId>j4ml</groupId>
         <artifactId>j4ml-neuroph</artifactId>
-        <version>0.9-SNAPSHOT</version>
+        <version>1.0</version>
     </dependency>
 
     <dependency>
         <groupId>j4ml</groupId>
         <artifactId>j4ml-clas12</artifactId>
-        <version>0.9-SNAPSHOT</version>
+        <version>1.0</version>
     </dependency>
 	
     <dependency>
--- a/etc/bankdefs/hipo4/dc.json
+++ b/etc/bankdefs/hipo4/dc.json
@@ -6,6 +6,7 @@
     "info": "reconstructed hits using DC wire positions",
     "entries": [
        {"name":"id",          "type":"S", "info":"id of the hit"},
+       {"name": "indexTDC",    "type":"S", "info":"inex in the bank DC::TDC"},
        {"name":"status",      "type":"S", "info":"id of the hit"},       
        {"name":"sector",      "type":"B", "info":"DC sector"},
        {"name":"superlayer",  "type":"B", "info":"DC superlayer (1...6)"},
--- a/reconstruction/dc/src/main/java/org/jlab/rec/dc/banks/HitReader.java
+++ b/reconstruction/dc/src/main/java/org/jlab/rec/dc/banks/HitReader.java
@@ -54,7 +54,8 @@ public class HitReader {
     private IndexedTable docares     = null;
     private IndexedTable time2dist   = null;
     private IndexedTable t0s         = null;
-
+    
+    private int numTDCBankRows = -1;
     private List<Hit> _DCHits;
     private List<FittedHit> _HBHits; //hit-based tracking hit information
     private List<FittedHit> _TBHits; //time-based tracking hit information
@@ -263,6 +264,7 @@ public class HitReader {
 
         RawDataBank bankFiltered = new RawDataBank(bankNames.getTdcBank(), rawBankOrders);
         bankFiltered.read(event);
+        this.set_NumTDCBankRows(bankFiltered.rows());
         for (int i = 0; i < bankFiltered.rows(); i++) {
             int sector     = bankFiltered.getByte("sector", i);
             int layer      = (bankFiltered.getByte("layer", i)-1)%6 + 1;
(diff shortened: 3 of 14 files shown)


zieglerv:
isExceptionalCluster and isExceptionalFittedCluster are basically copies of each other. 
It may be better to use only one method for the algorithm to decide if the cluster is exceptional and apply it to Hit or Fittedhit, something like:
/**
 * Check if one or more layers are skipped in the cluster
 * @param hitsInClus the hits in a cluster (can be either Hit or FittedHit)
 * @param nlayr the number of layers
 * @return true if one or more layers are skipped in the cluster
 */
private boolean isExceptionalClusterHelper(List<? extends Hit> hitsInClus, int nlayr) {
    // Initialize array to count hits in each layer
    int[] nlayers = new int[nlayr];

    // Count hits for each layer in a single pass through the hits
    for (Hit hit : hitsInClus) {
        int layer = hit.get_Layer() - 1;  //  layer numbering starts from 1
        if (layer >= 0 && layer < nlayr) {
            nlayers[layer]++;
        }
    }

    // Check for skipped layers (special cases for layer pairs)
    if ((nlayers[0] == 0 && nlayers[1] == 0) || (nlayers[4] == 0 && nlayers[5] == 0)) {
        return true;
    }

    // Check for skipped layers in the middle (layers 0 to 3)
    for (int l = 0; l < 4; l++) {
        if (nlayers[l] > 0 && nlayers[l + 1] == 0) {
            return true;
        }
    }

    return false;
}

/**
 * Wrapper for checking if a cluster of `Hit` objects is exceptional.
 */
public boolean isExceptionalCluster(List<Hit> hitsInClus) {
    return isExceptionalClusterHelper(hitsInClus, 6); // 6 layers for `Hit` objects
}

/**
 * Wrapper for checking if a cluster of `FittedHit` objects is exceptional.
 */
public boolean isExceptionalFittedCluster(List<FittedHit> hitsInClus) {
    return isExceptionalClusterHelper(hitsInClus, 6); // 6 layers for `FittedHit` objects
}

zieglerv:
A lot of 3-hit clusters are noise.  Which is why on the plot of the cluster size on page 11 of the document, there is a spike at 3.  What is the fraction of these 3-hit clusters on track after AI inference?  Also the problem with using clusters with only 3 layers is that the left-right ambiguity can be hard to resolve.  For example, on page 4, Exc1 it is 100 ambiguous unless it is in region3 which has the minstagger; Exc2 has ambiguities for the small docas, and it does not matter that much in this case.  It would be nice to include in this document a MC study of the left-right ambiguity accuracy as a function of cluster size for special cluster type1 (exc1)  and type2 (exc2).  
In the exceptional clusters you may want to add a constraint that the left-right ambiguity (LR)  is solvable.  For instance, for type1 clusters, you may want to require a double hit in a layer for regions 1 and 2 to pin down LR.
It may be better to toss out a 3-layer cluster with wrong left-right ambiguity and rely on 5 super layer tracking than to keep it and possibly bias the track.

zieglerv (dismissed):
See my comments in the conversation section about 3-hit clusters.  I suggest requiring a double hit in a layer for type 1 exceptional clusters (where the first or last 2 layers are lost) that are in region 1 or 2 to assist resolving the left-right ambiguity for those cases. In region 3 the mini stagger does that job.  

Suggestions for code tidiness: isExceptionalCluster and isExceptionalFittedCluster are basically copies of each other.
It may be better to use only one method for the algorithm to decide if the cluster is exceptional and apply it to Hit or Fittedhit, something like:
/**

Check if one or more layers are skipped in the cluster

@param hitsInClus the hits in a cluster (can be either Hit or FittedHit)

@param nlayr the number of layers

@return true if one or more layers are skipped in the cluster
*/
private boolean isExceptionalClusterHelper(List<? extends Hit> hitsInClus, int nlayr) {
// Initialize array to count hits in each layer
int[] nlayers = new int[nlayr];

// Count hits for each layer in a single pass through the hits
for (Hit hit : hitsInClus) {
int layer = hit.get_Layer() - 1; // layer numbering starts from 1
if (layer >= 0 && layer < nlayr) {
nlayers[layer]++;
}
}

// Check for skipped layers (special cases for layer pairs)
if ((nlayers[0] == 0 && nlayers[1] == 0) || (nlayers[4] == 0 && nlayers[5] == 0)) {
return true;
}

// Check for skipped layers in the middle (layers 0 to 3)
for (int l = 0; l < 4; l++) {
if (nlayers[l] > 0 && nlayers[l + 1] == 0) {
return true;
}
}

return false;
}

/**

Wrapper for checking if a cluster of Hit objects is exceptional.
*/
public boolean isExceptionalCluster(List hitsInClus) {
return isExceptionalClusterHelper(hitsInClus, 6); // 6 layers for Hit objects
}
/**

Wrapper for checking if a cluster of FittedHit objects is exceptional.
*/
public boolean isExceptionalFittedCluster(List hitsInClus) {
return isExceptionalClusterHelper(hitsInClus, 6); // 6 layers for FittedHit objects
}

tongtongcao:
Nice suggestion to use wrapper to combine two similar functions. 
Page 11 in the shared document shows distributions of clusters directly from DC clustering before AI prediction of cluster combo.
It is not surprise that there are plenty of low-size clusters. So we just looser cluster limit for so-called exceptional clusters.
Indeed, there is LR ambiguous for some of exceptional clusters. 
Most likely, low-size clusters have more chance to get LR ambiguous than high-size clusters.
Regards to LR ambiguity of clusters, we actually handle it during TB tracking by build two-folder clusters for a cluster with LR ambiguity

baltzell:
in today's meeting it sounds like the requested change is ok as is ?

zieglerv:
Ok with me.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________
From: Nathan Baltzell ***@***.***>
Sent: Thursday, January 30, 2025 5:53:36 PM
To: JeffersonLab/coatjava ***@***.***>
Cc: Veronique Ziegler ***@***.***>; Review requested ***@***.***>
Subject: [EXTERNAL] Re: [JeffersonLab/coatjava] DC clustering updates (PR #419)


in today's meeting it sounds like the requested change is ok as is ?

raffaelladevita:
Ok with me too but I would wait to merge PR440 and tag first a release before merging this


On Jan 30, 2025, at 5:55 PM, Veronique Ziegler ***@***.***> wrote:


Ok with me.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________
From: Nathan Baltzell ***@***.***>
Sent: Thursday, January 30, 2025 5:53:36 PM
To: JeffersonLab/coatjava ***@***.***>
Cc: Veronique Ziegler ***@***.***>; Review requested ***@***.***>
Subject: [EXTERNAL] Re: [JeffersonLab/coatjava] DC clustering updates (PR #419)


in today's meeting it sounds like the requested change is ok as is ?

baltzell (approved):

PR #432, opened 2025-01-14
FIx AHDC waveform length

baltzell:


etc/bankdefs/hipo4/data.json
```diff
@@ -10,27 +10,150 @@
...
+    "info": "ADC bank for the ALERT Wire Chamber",
+    "entries":[
+      { "name":"sector"    , "type":"B", "info":"sector 1"},
+      { "name":"layer"     , "type":"B", "info":"layer 1-5"},
+      { "name":"component" , "type":"S", "info":"wire number"},
+      { "name":"order"     , "type":"B", "info":"Stereo angle 0 - in, 1 - out"},
+      { "name":"ADC"       , "type":"I", "info":"ADC maximum"},
+      { "name":"time"      , "type":"F", "info":"time from pulse fit"},
+      { "name":"ped"       , "type":"S", "info":"pedestal from pulse analysis"},
+      { "name":"id"        , "type":"S", "info":"pointer to row in waveform bank"}
```
raffaelladevita: if it's the row number starting from 0, can we call it index?

raffaelladevita (changes requested):

raffaelladevita (approved):

PR #434, opened 2025-01-15
removed MC specific condition in DC hit reconstruction

raffaelladevita:


raffaelladevita:
Two MC-specific conditions still remained in FittedHit:
* the hit uncertainty is defined differently for data/mc;
* the time walk is not applied to mc.

The check is done here on the presence of the MC::Particle bank, which is compatible with using real run numbers. Further evaluation is needed to decide whether these can be removed

baltzell:
superceded by #441

PR #445, opened 2025-01-27
New features for ALERT Kalman filter+optimized parameters

efuchey:
* Optimized number of iterations and step size for numerical calculation of ddoca/dx:
https://clasweb.jlab.org/wiki/images/7/7b/Alert_progress_20250123.pdf
* Redid optimization after introduction of AI track finder;

* Added AHDC hit residuals (post-fit and pre-fit) in AHDC hit output: 
- residuals in the AHDC::Hits list in the alert.json file;
- filling the hits residuals in the RecoBankWriter;
- added residual and residual_prefit in ahdc/Hit/Hit.java
- added a identification flag to match ahdc/KalmanFilter/Hit.java to ahdc/Hit/Hit.java
- affect the track parameters to the KFTrack right after the fit;
- redo a forward indicators pass without correction for post-fit residual evaluation;

* Added optional use of vertex input to constraint "beam hit": 
- Added flag "vertexdefined" in KalmanFilter and KFitter to use vertex;
- added a different constructor for the list of backwards-indicators to include vertex info;
- Setting sigma_Z_beam_Hit^2 to 100mm2 if "vertexdefined" is true;

* Attempt to include hit "sign" / left-right disambiguation:
 - added "virtual wires" located at the distance-of-closest-approach of the actual wire, on each side of the wire;
 - added hit sign parameter in KalmanFilter/Hit class;
  added a new distance function to KalmanFilter/Hit class calculate the distance of a point to the correct virtual wire depending on the sign;
 - attempt to modify the "h" function to call new distance function


mathieuouillon (dismissed):
Look good for me.

baltzell (approved):

PR #446, opened 2025-01-27
Recover more data from AHDC pulse analysis

ftouchte:
News : 

- I have extended the `AHDC::adc`  bank structure to stored more data from pulse analysis

- I have added default values at some parameters of `ModeAHDC` extractor


whit2333:
What is the status of this PR?

baltzell:
modifying HipoExtractor like that is a nogo

whit2333:
> modifying HipoExtractor like that is a nogo

@baltzell I think you might need to expand on this.

baltzell:
Sorry Felix, yes, if it's customization for ALERT, it should probably get overriden in the existing class(es) containing ALERT customization.  As is, it looks like it breaks the other classes that extend the same class (org.jlab.detector.pulse.HipoExtractor).

And I think the method in question even mentions that in its comments:  https://github.com/JeffersonLab/coatjava/blob/96e1e862edb1252e1ef7aa8466c03edb1d4d2645/common-tools/clas-detector/src/main/java/org/jlab/detector/pulse/HipoExtractor.java#L48

ftouchte:
Hi Nathan, sorry if I didn't understand : so I can override the update method in [ModeAHDC.java](https://github.com/JeffersonLab/coatjava/blob/96e1e862edb1252e1ef7aa8466c03edb1d4d2645/common-tools/clas-detector/src/main/java/org/jlab/detector/pulse/ModeAHDC.java), right ?

ftouchte:
@baltzell Is it ok ? To override the update method I needed to use some private methods of HipoExtractor (copyIndices and getPulses). So I had either to copy these methods in ModeAHDC or declare them accessible for all class of the same package (by commenting the keyword `private`). What is better ?

whit2333:
> override the update method I needed to use some private methods of HipoExtractor (

@ftouchte @baltzell, this is a good point. It seems if they are useful for subclasses they should be marked `protected` instead of `private`.

 @ftouchte, I suggest changing the `HipoExtractor` methods you want to use as `protected`.

common-tools/clas-detector/src/main/java/org/jlab/detector/pulse/HipoExtractor.java
```diff
@@ -91,15 +91,15 @@ protected void update(int n, IndexedTable it, Bank wfBank, Bank adcBank) {
         }
     }
 
-    private static void copyIndices(Bank src, Bank dest, int isrc, int idest) {
+    /*private*/ static void copyIndices(Bank src, Bank dest, int isrc, int idest) {
```
whit2333: `protected` is what you want here:
https://www.w3schools.com/java/ref_keyword_protected.asp

whit2333 (dismissed):

common-tools/clas-detector/src/main/java/org/jlab/detector/pulse/HipoExtractor.java
ftouchte: do I still have to declare them `protected` since you have just approved the changes ?

whit2333: Please.

whit2333 (approved):

PR #448, opened 2025-01-29
feat: tell `dependabot` to check `clasweb` maven repos too

c-dilks:
If we can still rely on `clasweb` availability, we might as well check its maven repositories for dependency updates too.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,9 +1,21 @@
 version: 2
+
+registries:
+  clas12maven:
+    type: maven-repository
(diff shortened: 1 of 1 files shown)


c-dilks:
Can't really test this til it's merged, but I just did a similar change in `clas12-timeline` (https://github.com/JeffersonLab/clas12-timeline/pull/270), and dependabot's logs indicate that it tried to search `clas12maven` (but got 404 error on the only updatable clas12maven dependency, due to lack of a `maven-metadata.xml` file).

baltzell (approved):

PR #452, opened 2025-02-02
version bump

baltzell:
(still need to automate this)

### DIFF
--- a/install-clara
+++ b/install-clara
@@ -3,7 +3,7 @@
 # Default versions:
 grapes=2.17
 clara=5.0.2
-coatjava=11.1.0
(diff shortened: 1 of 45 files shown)


install-clara
```diff
@@ -3,7 +3,7 @@
 # Default versions:
 grapes=2.17
 clara=5.0.2
-coatjava=11.1.0
+coatjava=11.1.1
```
c-dilks, line 6: Why is this one behind?

baltzell: it's not, just installing the latest release

c-dilks (approved):

PR #458, opened 2025-02-04
ci: run more `eb` tests

c-dilks:


.github/workflows/maven.yml
```diff
@@ -71,25 +71,37 @@ jobs:
   test_coatjava:
     needs: [ build ]
     strategy:
-      fail-fast: true
+      fail-fast: false
```
c-dilks, line 74: consider reverting this before merging (currently `false` to see which tests fail)
```suggestion
      fail-fast: true
```

baltzell:
I regenerated the electronFTpion/electronFTproton/electronpionC test files, new random number seeds, rather than change the thresholds.

Issue #475, opened 2025-02-17
TI timestamp mismatch tolerance


baltzell:
Sergey suggests setting it to 5 tics.  Currently it's 1(?).

whit2333:
What is the value of the [`RUN::config` timestamp](https://github.com/JeffersonLab/coatjava/blob/development/reconstruction/tof/src/main/java/org/jlab/service/ctof/CTOFEngine.java#L100) expected to be?

baltzell:
I think it's a "global" TI timestamp, which is maybe only well-defined when all TI timestamps are equal.  I think it's only used for (TDC??) jitter corrections.

baltzell:
Partially closed by #482, but later realized ...

We must add an additional requirement that the `i0` reference (which is what goes in `RUN::config.timestamp` if "all" ROCs have "equal" TI timestamps) is not from ATOF (same as is done currently for the `tiMaster`, which is defined as the one that has the "HEAD" bank [0xe10f](https://github.com/JeffersonLab/clas12-parms/blob/b5851869891a4652ce51363af1e1b4fa783f4e8b/clonbanks.xml#L257-L268)/[57615](https://github.com/JeffersonLab/coatjava/blob/20f3d25b09e1c5f40b2f42a3aa8696040144595d/common-tools/clas-detector/src/main/java/org/jlab/detector/decode/CodaEventDecoder.java#L237-L252))

Then, #482 would work as intended for all data sets until we choose a better approach.

PR #476, opened 2025-02-18
Added more "clicks" to timestamp tolerance.

whit2333:
Addresses issue #475.

I just added the 5 clicks that Sergey mentioned. To do this right, we need to know if the detector is using a TI board or TIpcie card. But I have verified this works with a recent "fixed" run 20764 where I can see the CTOF reconstruction.

Tests: 
- [no more errors in decoder](https://code.jlab.org/hallb/alert/c12/-/jobs/5862#L318)
- [no mismatches from CTOF in reconstruction](https://code.jlab.org/hallb/alert/c12/-/jobs/5866#L61)

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -7,6 +7,9 @@ workflow:
     - if: '$CI_COMMIT_BRANCH == "development"'
     - if: '$CI_COMMIT_TAG'
 
+variables:
+  KUBERNETES_MEMORY_LIMIT: "8Gi"
+
 default:
   image: ubuntu:noble
   retry: 2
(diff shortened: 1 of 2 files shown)


whit2333:
test works: https://code.jlab.org/hallb/alert/c12/-/jobs/6237#L306

baltzell (dismissed):

baltzell:
Have we used 78 before in CLAS12?

whit2333:
@baltzell, Sorry I should mention that 78 is the ATOF. 
I have added a comment reflecting this.

baltzell:
Right, my concern is whether we've had a 78 in (production) data before.  If not, we can do it like this at least temporarily.

baltzell:
How about trigger on detector type instead crate number?   It's easy, accessible from the same object, and avoid conflicts from reused crate numbers.  We also need to change the variable names for clarity, e.g., "i0" which is probably more correctly "iMaster".

whit2333:
use https://github.com/JeffersonLab/coatjava/pull/482

PR #482, opened 2025-02-21
increase timestamp tolerance for ATOF

baltzell:
alternative to #476


whit2333 (approved):
This is more clear than  the ROC number.

PR #486, opened 2025-02-24
Iss485

tongtongcao:
New AI model plans to use local y at layers 1&6 for a cluster, calculated by a linear fitting function of the cluster in LC, as features for cluster combo prediction.
The features need to be added into DC cluster bank.
The branch was made when updates for DC clustering.  Most of following commits are for updates for DC clustering, which have been merged into development.


reconstruction/dc/src/main/java/org/jlab/rec/dc/cluster/ClusterCleanerUtilities.java
zieglerv, line 1: Instead of having 2 methods for exceptional clusters that do the same thing, perhaps something like this to avoid code copy and paste: 
/**
 * Common logic for checking if one or more layers are skipped in a cluster.
 * 
 * @param hitsInClus the hits in a cluster
 * @param nlayr the number of layers in the detector
 * @return true if one or more layers are skipped in the cluster, false otherwise
 */
private boolean hasSkippedLayer(List<? extends Hit> hitsInClus, int nlayr) {
    // Count hits in each layer
    int[] nlayers = new int[nlayr];
    
    // Initialize all layers hit count to 0
    for (int i = 0; i < nlayr; i++) {
        nlayers[i] = 0;
    }

    // Count hits for each layer
    for (Hit hit : hitsInClus) {
        int layer = hit.get_Layer() - 1; // Adjust for 0-indexed array
        if (layer >= 0 && layer < nlayr) {
            nlayers[layer]++;
        }
    }

    // Check if layers are skipped
    if ((nlayers[0] == 0 && nlayers[1] == 0) || (nlayers[4] == 0 && nlayers[5] == 0)) {
        return true;
    }

    for (int i = 0; i < 4; i++) {
        if (nlayers[i] > 0 && nlayers[i + 1] == 0) {
            return true;
        }
    }

    return false;
}

/**
 * Checks if one or more layers are skipped in the cluster of normal hits.
 *
 * @param hitsInClus the hits in a cluster
 * @return true if one or more layers are skipped in the cluster, false otherwise
 */
public boolean isExceptionalCluster(List<Hit> hitsInClus) {
    return hasSkippedLayer(hitsInClus, 6);
}

/**
 * Checks if one or more layers are skipped in the cluster of fitted hits.
 *
 * @param hitsInClus the hits in a cluster
 * @return true if one or more layers are skipped in the cluster, false otherwise
 */
public boolean isExceptionalFittedCluster(List<FittedHit> hitsInClus) {
    return hasSkippedLayer(hitsInClus, 6);
}

zieglerv (changes requested):
Code logic suggestion

tongtongcao:
Code modification, suggested by Veronique, has been taken at another PR, which has been merged into development.

baltzell:
Should this be closed?

tongtongcao:
The PR does not need to be merged into development now, so it could be closed. I will keep it in my local.

PR #494, opened 2025-02-27
DC v2: includes TFLight fix and new dc beta time walk with constants from /calibration/dc/v2

raffaelladevita:


reconstruction/dc/src/main/java/org/jlab/rec/dc/banks/HitReader.java
raffaelladevita, line 1: why is the old time2dist table still being used?

zieglerv (approved):

tongtongcao (approved):
All looks fine.

hauenst (approved):
All good

PR #499, opened 2025-03-12
Update DC reconstruction at the denoisinng level

tongtongcao:
There are two updates in this PR:
1. Cancel SNR: from attached files, you can see that there is almost no difference with and without application of SNR after changing parameter for limits of missing layers from 2 to 3.
2. Cancel limit for total of raw DC hits. Instead, set limit for DC hits after AI-denoising.
Here are several slides for discussion of updates: 
[iss471_1.pdf](https://github.com/user-attachments/files/19215646/iss471_1.pdf)


zieglerv (dismissed):

reconstruction/dc/src/main/java/org/jlab/rec/dc/banks/HitReader.java
```diff
@@ -205,38 +205,24 @@ private int getJitter(int sector, int layer, int wire, int order) {
...
-                             Clas12NoiseResult results) {
-
+    public void fetch_DCHits(DataEvent event) {
+        this.initialize(event);
+        
         _DCHits = new ArrayList<>();
-
-        IndexedList<Boolean> noise = new IndexedList<>(4);
         
         RawDataBank bankDGTZ = new RawDataBank(bankNames.getTdcBank(), OrderGroups.NODENOISE);
```
raffaelladevita: This, up to line 239, is no more used and can be removed

raffaelladevita (changes requested):

reconstruction/dc/src/main/java/org/jlab/rec/dc/banks/HitReader.java
tongtongcao: Done.

raffaelladevita (approved):

PR #502, opened 2025-03-12
Add skip option

### LINKED ISSUE #501: Add a -skip option to the decoder
currently to temporarily workaround changes to new helicity decoder board firmware

baltzell:
closes #501

### DIFF
--- a/common-tools/clas-detector/src/main/java/org/jlab/detector/decode/CLASDecoder4.java
+++ b/common-tools/clas-detector/src/main/java/org/jlab/detector/decode/CLASDecoder4.java
@@ -767,6 +767,7 @@ public class CLASDecoder4 {
         parser.addOption("-s", "0.5","solenoid current in the header bank");
         parser.addOption("-x", null,"CCDB timestamp (MM/DD/YYYY-HH:MM:SS)");
         parser.addOption("-v","default","CCDB variation");
+        parser.addOption("-skip","0", "number of events to skip");
 
         parser.parse(args);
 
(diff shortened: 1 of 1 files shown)


common-tools/clas-detector/src/main/java/org/jlab/detector/decode/CLASDecoder4.java
```diff
@@ -767,6 +767,7 @@ public static void main(String[] args){
         parser.addOption("-s", "0.5","solenoid current in the header bank");
         parser.addOption("-x", null,"CCDB timestamp (MM/DD/YYYY-HH:MM:SS)");
         parser.addOption("-v","default","CCDB variation");
+        parser.addOption("-skip","0", "number of events to skip");
```
raffaelladevita, line 770: I think the parser doesn't support using strings that contain one of the other strings (in this case -skip contains -s). Did you check -s still works?

baltzell: will do ... meanwhile, note it's based on a Map<String,String> for option name->value (and the usual containsKey/get for access)

baltzell:
looks like this happens before the map storage/lookup: 

 https://github.com/JeffersonLab/coatjava/blob/f1d8a2c7b6e5d13602b3c7902d1e5bf1bbb03740/common-tools/clas-utils/src/main/java/org/jlab/utils/options/OptionValue.java#L42

PR #506, opened 2025-03-24
Retain and pass additional timing info for DREAM

baltzell:


baltzell:
confirmed it works on 21046

raffaelladevita (approved):

PR #517, opened 2025-03-28
fix: remove and ignore Eclipse JDT language server files

c-dilks:
Ignore and remove dotfiles from `jdtls` (https://github.com/eclipse-jdtls/eclipse.jdt.ls)


c-dilks:
Although these may be useful to have,
- not everyone uses this LSP
- the dot files aren't provided for _all_ modules
- they typically get regenerated, _e.g._, by an IDE that uses `jdtls`
- they may fall out of date

If we _really_ want these, make a pre-commit hook...

baltzell (approved):

PR #518, opened 2025-03-29
Do not use the MVTFitter in AHDC and take into account the time of the fisrt channel

ftouchte:


baltzell:
Felix, let's get your branch up to date, and write only your new commits, so it's easier to see what's going on here.  One thing that helps there is to always work on a new branch, not the same one you want to ultimately merge to.

ftouchte:
I close this pull request. I will open a new one. (Because I re-forked coatjava.)

ftouchte:
PR closed.

PR #525, opened 2025-04-01
Fixes on Kalman Filter: address issue #553

efuchey:
Looping on all track candidates in Kalman filter instead of the first track only, to address:
https://github.com/JeffersonLab/coatjava/issues/553

Added simulation flag self-configuration in AHDC Engine processing;
Initialized AHDC Kalman Filter vertex constraint to 0 to avoid a nonsensical value if the vertex constraint is enabled but no vertex constraint is available.
Ensured this constraint will not be fetched from MC banks if MC banks are not available.


efuchey:
Closing pull request to redo it from a different branch

PR #529, opened 2025-04-03
Alert engine for combined ATOF/AHDC reconstruction

whit2333:
- Moved ATOF Track projections to ALERTEngine since it will need a track
- Added many todo comments in code.
- Cleaned up various if statements for clarity.
- moved services to proper project directory structure that is picked up by javadocs


baltzell:
I suspect people might prefer to leave the old ones intact for now, maybe just suffix their names to designate they're the old ones.

whit2333:
> I suspect people might prefer to leave the old ones intact for now, maybe just suffix their names to designate they're the old ones.

Not sure what you are saying? The old Engines are  still there and used. The only change is I moved the Track projections into ALERTEngine where it is more appropriate.

baltzell:
Yes, the old ones' *names* are still there and used, but their functionality is changed, and their previous functionality is no longer easily accessible.  In the short term, it'd probably be better to keep their functionality easily accessible, and one easy way to do that is by renaming those old engine classes with a suffix, rather than replacing them, simultaneously designating that they'll be going away in the near future.

whit2333:
I still don't understand. All I have done is add a new engine called ALERTEngine. None of the functionality changed. I can move the projection back to ATOFEngine but the next PR will move it out of there....

baltzell (dismissed):

baltzell (approved):

PR #515, opened 2025-03-28
build!: various cleanups

c-dilks:
> [!IMPORTANT]
> **If you are a developer, after this PR is merged, please use the `--clean` option when you run `build-coatjava.sh` for the first time! This will prevent old JAR files from creeping into your new build.**

## `build-coatjava.sh` updates
- now _optionally_ runs `mvn clean`, with the new option `--clean`
  - running `mvn clean` every time was the previous default behavior, but only for `common-tools/coat-lib` POM
  - optionally running `mvn clean` is more in line with other build automation tools
- installation prefix now uses an absolute path rather than relative path `./coatjava`; this should prevent the user from accidentally removing some _other_ directory named `coatjava`

## `version-bump.sh` updates
- no more need to specify the old version number, _e.g._:
```bash
./version-bump.sh 13.0.0
```
- now uses `mvn` to directly update the project version, rather than `sed`
    - `sed` is still needed for hard-coded version numbers in a couple scripts
    - version number in `install-clara` is now bumped
    - specifying the version number in one place, such as a parent POM property, is [practically a no-go in Maven](https://stackoverflow.com/questions/19123013/maven-version-with-a-property); using a `mvn` command is the preferred way
- automatically creates a new `git` branch and runs `git commit`; the user may then run `git push` and open a PR
    - requires the user to start from the `development` branch
    - add the argument `--no-git` to disable this `git` automation

## POM changes
- Combined `common-tools/coat-lib` POM with `common-tools` POM
    - `common-tools/coat-lib` is now just another module, instead of independent
    - removed unused `common-tools/parent/pom.xml`
    - removed unused `common-tools/build.sh`, and `common-tools/README.md`, its only reference
- POM project version numbers are now up-to-date _everywhere_
    - this is a result of using `mvn` instead of `sed` to bump the version
    - for many JARs, we had a lot of 1.0 and 1.1 snapshot versions, and it appears their version numbers were _never_ bumped
    - some POM modules also now use the up-to-date version number, _e.g._, `common-tools/cnuphys/magfield`
- see **comments below** for comparisons of the installation tree and JAR files

PR #538, opened 2025-04-07
build!: use same parent POM for all modules, and fix `javadoc` inter-module linking

### LINKED ISSUE #536: Javadoc multi project/subproject strucuture
Building the javadocs has made clear there is no utility in managing what is effectively many subprojects (mvn's modules). There roughly 17 library modules and >20 detector reconstruction modules. In javadocs, there appears to be no linking between these. For example, [band engine](https://jeffersonlab.github.io/coatjava/javadoc/reconstruction/band/org/jlab/service/band/BANDEngine.html) does not link to `org.jlab.clas.reco.ReconstructionEngine`.

Perhaps I am missing the benefits of having these divisions...

c-dilks:
> [!IMPORTANT]
> The front page of the API documentation will look quite different
> - If you're confused by this, try 'SEARCH', 'INDEX', or 'TREE' at the top of the page
> - Many links are now being auto-generated, which should make the documentation significantly easier to navigate

- follows up on #515 by making `parent/pom.xml` be _the_ parent POM for _everyone_
  - only `common-tools/cnuphys` is affected, since those modules were using `common-tools/cnuphys/parent/pom.xml` as their parent POM, which was pretty much a "subset" of `parent/pom.xml`
  - removed a couple more unused `pom.xml` files (in `common-tools/cnuphys`)
  - adds XML preamble to all `pom.xml` files that lacked it

- fixes #536 by aggregating `javadoc` results
  - many more links are now being auto-generated
  - the front "overview" page is now generated by `javadoc`, instead of some HTML generation
  - the version number is now included in the documentation webpage title

- `build-javadocs.sh` is :warning: **removed** and replaced by `build-coatjava.sh --docs`
  - documentation is now "installed" to `coatjava/share/doc/coatjava/html/`, which is more consistent with the Filesystem Hierarchy Standard (`man hier`)
    - previous location `docs/javadoc` is now fully `git` ignored
  - CI configs updated accordingly; modified GitLab config is :warning: **not tested** with this PR, but should be trivial to fix if it doesn't work


baltzell (dismissed):

c-dilks:
Here's a sample deployment from my fork: https://c-dilks.github.io/coatjava/javadoc/index.html

c-dilks:
Just noticed that the Javadoc webpage size drops from ~160 MB down to ~9 MB. This is entirely because all the font resources are duplicated _multiple_ times.

baltzell:
After this, we can do #612

baltzell (approved):

PR #544, opened 2025-04-08
Change geometry for the hits to match reality. (90 shift + conterclock)

mathieuouillon:


baltzell:
If I understand correctly, the proposal is to make the geometry service s/l/c/o convention match the tracking code rather than the expected/ideal geometry?

raffaelladevita:
Mathieu can confirm but I think this PR is updating the AHDC geometry to account for a change in the detector wire numbering specifications that happened some time ago and was not propagated to the software:
- the current geometry assumes the first wire of each layer is at phi=-90 deg and the wire number decreases for increasing phi;
- with this PR, wire 1 is at phi=0, and the wire number increases with phi.

raffaelladevita (approved):

Issue #553, opened 2025-04-09
A loop over the tracks is missing in the Kalman Filter of AHDC


ftouchte:
https://github.com/JeffersonLab/coatjava/blob/4969105ec314f44d5afb9d1c1d72c791d850b645/reconstruction/alert/src/main/java/org/jlab/rec/ahdc/KalmanFilter/KalmanFilter.java#L40

When they exist, only the attributes of the first track are computed.

[Here](https://github.com/JeffersonLab/coatjava/blob/4969105ec314f44d5afb9d1c1d72c791d850b645/reconstruction/alert/src/main/java/org/jlab/rec/ahdc/KalmanFilter/KalmanFilter.java#L185C4-L185C53) for example, we only set position and momentum for the first track of the list. I think each reference to `tracks.get(0)` should be replace by `track`.

```java
private void propagation(ArrayList<Track> tracks, DataEvent event, boolean IsMC) {
        for (Track track : tracks) {
                  // replace all tracks.get(0) by track
        }
}
``` 

![Image](https://github.com/user-attachments/assets/15964830-ca0a-4e57-9d1d-d29d9c9803e2)

baltzell:
closed by #555

PR #555, opened 2025-04-09
Fixes on Kalman Filter: address issue #553 (redo)

efuchey:
Looping on all track candidates in Kalman filter instead of the firs track only, to address:

https://github.com/JeffersonLab/coatjava/issues/553

Added simulation flag self-configuration in AHDC Engine processing; Initialized AHDC Kalman Filter vertex constraint to 0 to avoid a nonsensical value if the vertex constraint is enabled but no vertex constraint is available. Ensured this constraint will not be fetched from MC banks if MC banks are not available.

### DIFF
--- a/reconstruction/alert/src/main/java/org/jlab/rec/ahdc/KalmanFilter/KalmanFilter.java
+++ b/reconstruction/alert/src/main/java/org/jlab/rec/ahdc/KalmanFilter/KalmanFilter.java
@@ -39,7 +39,7 @@ public class KalmanFilter {
 	private void propagation(ArrayList<Track> tracks, DataEvent event, boolean IsMC) {
 
 		try {
-			double vz_constraint;
+			double vz_constraint = 0;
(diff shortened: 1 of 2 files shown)


raffaelladevita (commented):
Do tracks have ay overlap, i.e. share some of the hits?

ftouchte:
@raffaelladevita For now, I cannot answer your question, may be @efuchey can. But, this PR fixes this issue [#553](https://github.com/JeffersonLab/coatjava/issues/553). 

I need this update to complete another work. I'm working on an extension of the AHDC::Track and AHDC::KFTrack and AHDC::Hits banks to store the **trackId** and other quantities. By comparing the trackId and the number of hits in each track, I will be able to answer this question. [extend/ahdc-track](https://github.com/JeffersonLab/coatjava/compare/development...ftouchte:coatjava:extend/ahdc-tracks)

ftouchte (approved):
It's ok for me.

Issue #493, opened 2025-02-27
AHDC and ATOF bank and variable names should be consistent


Sangbaek:
## Troubles

In [alert.json](https://github.com/JeffersonLab/coatjava/blob/development/etc/bankdefs/hipo4/alert.json), the bank and variable names would be better to follow the consistent capitalization.

- AHDC::Clusters and ATOF::clusters are similar banks, but differ in capitalization
- AHDC::Hits and ATOF::hits are similar banks, but differ in capitalization
- AHDC::Hits:ID and ATOF::hits:id are similar variables but differ in capitalization
- AHDC::Hits:ID and AHDC::Hits:Doca start with upper cases, but AHDC::Hits:layer, AHDC::Hits:superlayer, AHDC::Hits:wire start with lower cases.

## Why this is a problem?

The inconsistency in naming rules can induce errors in code, especially when collaborators besides who named these banks need to develop some codes, or analyze the data.

## Suggestions

All bank names should follow the rule: [Uppercase]:[Lowercase]:[Lowercase]

e.g.) AHDC::hits:x, to be consitent with most of bankdefs.

baltzell:
If this is going to happen, it needs to be done before requesting access to these banks in CED.

Sangbaek:
@ftouchte , just checking, is this still on your to-do list?

ftouchte:
Yes, it is. I will submit a pull request very soon.

c-dilks:
Seems like work on #558 is still ongoing; this issue will auto-close when it's merged.

PR #557, opened 2025-04-10
Extend AHDC tracks

ftouchte:


baltzell:
Since you're modifying some of the associated banks, it would be a good time to also address #493 (otherwise we should drop it by now).

ftouchte:
I could but this would also break some existing codes. If everyone is aware, I can do the modifications  in another pull request.

ftouchte:
Looking at data, it seems that several tracks can share the same hit. I will ask Raphael if it is an issue. May be what I defined is not a relevant track id. I will come back when the PR is ready.

baltzell (approved):

PR #561, opened 2025-04-11
feat: heap size control for `recon-util`

c-dilks:
@gavalian says it's too small; this is for testing a high-stats run in `clas12-validation` with larger max heap size.

### DIFF
--- a/bin/recon-util
+++ b/bin/recon-util
@@ -4,7 +4,7 @@
 
 export MALLOC_ARENA_MAX=1
 
-java -Xmx1536m -Xms1024m -XX:+UseSerialGC \
+java -Xmx3072m -Xms1024m -XX:+UseSerialGC \
     -cp "$CLAS12DIR/lib/clas/*:$CLAS12DIR/lib/services/*:$CLAS12DIR/lib/utils/*" \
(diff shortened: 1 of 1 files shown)


c-dilks:
testing with https://github.com/JeffersonLab/clas12-validation/pull/177

we know 1500m fails because https://github.com/JeffersonLab/clas12-validation/issues/172

c-dilks:
allocating 3GB speeds `clas12-validation` reconstruction up quite a bit: https://github.com/JeffersonLab/clas12-validation/pull/177

So.... let's make heap allocation control a feature, and leave the default the same.

baltzell:
Let's pursue a more generic approach that supports passing any arbitrary JVM options, for all these frontend scripts.  Could push it to OptionParser, or use the old-school `--` separator for CLI arguments in the shell wrapers, or ....

See #604

c-dilks:
> Let's pursue a more generic approach that supports passing any arbitrary JVM options, for all these frontend scripts. Could push it to OptionParser, or use the old-school `--` separator for CLI arguments in the shell wrapers, or ....
> 
> See #604

Fully agree, as discussed. Comment linked to #604 description.

PR #562, opened 2025-04-12
Revert "fix: geometry ..." and fix AHDC geometry

RDupre:
This reverts commit 1dfa2233c689ce76e3edb9690db0ed57bbc4933b.

This changes everything by 180 degrees and change the direction of rotation as needed per Raffaella's analysis


baltzell:
There's a at least a couple more commits that would be involved in reverting the geometry.  It might be best to reproduce https://logbooks.jlab.org/entry/4353923 with 12.0.1t before bothering.

baltzell:
I think #563 is what we want, just one more commit.

RDupre:
I am not reverting the two other commits that touch geometry, because I am modifying them heavily to get this. I do not understand why this pull request is closed. #563  will still have the 90 degrees offset.

PR #563, opened 2025-04-12
Revert AHDC geometry to original, as-designed

baltzell:


ftouchte:
I have a few comments. The analysis of Raffaella has been done using the last release **12.0.3t**, what is great because that limits the impact of the followings. 

I just notice that one of most [recent commits](https://github.com/JeffersonLab/coatjava/commit/b1d6e4c38d54faa934c2aede88b2e7656f6c301b#diff-53814f6c4e62686fb1a2c898359c98cfe20076b1a7f58bf0bfd96b8853d3ff82R73) invert the x and y position of the wire in the reconstruction service (it is after the release **12.0.3t**). That means that from here, there is a mismatch between the reconstruction service and the geometry service. As we plan to fix all AHDC related issues, can we prevent people to use hard coded definitions. I mean, everyone should refers to the geometry service. In that sense, the way the x and y positions are defined in `reconstruction/alert/src/main/java/org/jlab/rec/ahdc/Hit/Hit.java` and `reconstruction/alert/src/main/java/org/jlab/rec/ahdc/KalmanFilter/Hit.java` should be improved.

Unfortunatelly, all data cooked with the recent development branch (not the last release) should take into account my x and y inversion. :disappointed: (When I think, it is even more complicated because the Kalman Filter track may remain "true" because it uses a re-definition of Hits but  the standard track is "distorted").

@baltzell @raffaelladevita @RDupre @N-Plx @mathieuouillon @Sangbaek

PR #567, opened 2025-04-14
Undo unwanted modifications in rec/ahdc/Hit/Hit.java

ftouchte:
The issue comes from this commit "Add new entries in AHDC::Track, KFTrack and Hits". From here, we have a mismatch between the geometry service and the reconstruction service.

This will make all the AHDC geometry issues at the stage they were in the release **12.0.3t**.


baltzell (approved):

baltzell:
Is this compatible with #569, or needs updating?

ftouchte:
This can be closed. The error has been corrected in #569

Issue #560, opened 2025-04-11
Alert atof geometry : G4GenericTrap


maureeungaro:
The Alert ATOF G4GenericTrap volumes give these warnings:

*** G4Exception : GeomSolids1001
      issued by : G4GenericTrap::CheckOrder()
Vertices must be defined in clockwise XY planes - sector9_superlayer1_layer3_paddle8Re-ordering...

*** This is just a warning message. ***
-------- WWWW -------- G4Exception-END --------- WWWW -------

baltzell:
should be fixed after #569 and/or #570

baltzell:
closed by #569 and #571

PR #569, opened 2025-04-15
Ahdc geo revert and redo

raffaelladevita:
- Revert previous modification
- Correct AHDC wire positions to match real data with wire 1 in each layer at phi = 0 and wire number increasing with phi
- Shift AHDC center at z = 0
- Set the hit geometry info based on the geometry service (to be checked)

Geometry tested via elastic analysis of run 21342
Plot_04-14-2025_11 08 59_PM


baltzell:
#560's related?

RDupre:
#560 is the same issue we had when we inverted the phi in the AHDC. 
The fixes from this pull request look great, I will close mine.

mathieuouillon (approved):
Look good for me

PR #570, opened 2025-04-15
ahdc wire numbering maybe final

raffaelladevita:
consistent with https://code.jlab.org/hallb/alert/ahdc/ahdc_docs/-/blob/main/AHDC_number.pdf?ref_type=heads


raffaelladevita:
Checked once again with data 
Plot_04-15-2025_06 16 24_PM

Requires updating current TT to 
[ahdc.txt](https://github.com/user-attachments/files/19766285/ahdc.txt)
where the last two columns are the x and y of the wires at the upstream end

RDupre:
I am fine with the new numbering, wires of each layers are properly aligned. 

I am not sure about the mapping file associated with it. It seems to me it was generated without the cable swaps included (I assume this swap was not an artifact linked with the geometry service and that it should still be included). Was it included in a different way to produce the results presented here? 

I also think there is a slight error with the numbering due to the fact that the new definition does not give consistent shift in a given super-layer (by my calculation in SL1, L0 is moved by 28 wires (29 becomes 1), but L1 is only moved by 28). This is due to the change of direction in phi in the count method, the wires not aligned at 0 are not shifted by 180 degrees but less.

I attach what I get from my calculation: [AHDCmap_vn.txt](https://github.com/user-attachments/files/19773050/AHDCmap_vn.txt)

RDupre (approved):

baltzell:
replaced by #571

PR #572, opened 2025-04-16
Load raw hit cuts of ahdc from ccdb

ftouchte:
- load raw hit cuts from ccdb
- update the key value of the atof ccdb


RDupre:
Why do we have all these changes on the ATOF stuff? Are you sure you are not carrying some modifications unrelated to the pull request, or reverting another commit you have not implemented yet?

ftouchte:
The change on the ATOF is wanted. It is something I wanted to change a long time ago. Before, I defined the same key as for the AHDC without knowing the range of the (sector, layer, component, order) in the ATOF.

baltzell (approved):

PR #586, opened 2025-04-21
use logger

### LINKED ISSUE #580: AHDC helix fitter printing error
The AHDC service prints an error whenever there are not enough point for the helix track fitter:

[https://github.com/JeffersonLab/coatjava/blob/f772c46c36298c85246c30528f7bda11da2b[…]lert/src/main/java/org/jlab/rec/ahdc/HelixFit/HelixFitJava.java](https://github.com/JeffersonLab/coatjava/blob/f772c46c36298c85246c30528f7bda11da2b06f0/reconstruction/alert/src/main/java/org/jlab/rec/ahdc/HelixFit/HelixFitJava.java#L242)

This has been a problem cooking data, creating very heavy log files that fill up the log directory for some runs where this was printed every event. 

A temporary solution could be to simply remove printing that line.

But it also needs to be checked why there are no good event, and how to deal with them. And if it is worth cooking with the AHDC service, if other information from the track finding or KF can be still looked at.

@ftouchte @mathieuouillon Do you have insights please?

baltzell:
probably closes #580

### DIFF
--- a/reconstruction/alert/src/main/java/org/jlab/rec/ahdc/HelixFit/HelixFitJava.java
+++ b/reconstruction/alert/src/main/java/org/jlab/rec/ahdc/HelixFit/HelixFitJava.java
@@ -1,10 +1,15 @@
 package org.jlab.rec.ahdc.HelixFit;
 
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
 /** Helix Fit.
  *
  * This appears to be some code translation.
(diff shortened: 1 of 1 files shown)


efuchey:
Compiles, runs, removes the message. Approved.

efuchey (approved):
Compiles, runs, removes the message. Approved.

PR #590, opened 2025-04-21
Use CVMFS at build time too

baltzell:


c-dilks:
Looks good to me, but I think we can scope to the whole workflow rather than to specific steps: https://github.com/JeffersonLab/coatjava/pull/591

c-dilks:
I'm doing this to validation too, since that's also broken because of "too many connections": https://github.com/JeffersonLab/clas12-validation/pull/182

PR #610, opened 2025-05-02
version bump

baltzell:


c-dilks:
Rebasing, to see if that'll get rid of the pre-commit job attempt (if not, we'll disable it)

c-dilks:
> Rebasing, to see if that'll get rid of the pre-commit job attempt (if not, we'll disable it)

Nope, gotta be done in the settings.

raffaelladevita (approved):

PR #617, opened 2025-05-09
honor branch/tag in clara installer

baltzell:


install-clara
```diff
@@ -58,12 +58,8 @@ function cleanup() {
...
 
 function build_clara {
-    export CLARA_HOME=$1
-    git clone https://code.jlab.org/hallb/clas12/clara-java
-    cd clara-java
-    git checkout java21
-    ./gradlew && ./gradlew deploy
-    cd -
+    git clone --branch $clara --depth 1 https://code.jlab.org/hallb/clas12/clara-java
+    cd clara-java && CLARA_HOME=$1 ./gradlew deploy && cd -
```
c-dilks, line 62: Is `$CLARA_HOME` needed anywhere else? I'm asking since this change reduces the scope of `$CLARA_HOME` from the environment to this line.

baltzell: Yes, this instance of CLARA_HOME is only a build-time variable.

c-dilks (approved):

PR #624, opened 2025-05-10
fix versions

baltzell:


c-dilks:
There may be more of these cases lurking in older PRs / branches...

PR #625, opened 2025-05-11
build: enable `reconstruction/recoil` module and update its version

c-dilks:
Seems like we have not been building this one, possibly since v11.

### DIFF
--- a/reconstruction/pom.xml
+++ b/reconstruction/pom.xml
@@ -37,5 +37,6 @@
     <module>alert</module>
     <module>bg</module>
     <module>postproc</module>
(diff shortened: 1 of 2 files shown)


c-dilks:
> Seems like we have not been building this one, possibly since v11.

ah, we have _never_ been building it (at least as part of the _full_ `coatjava`): https://github.com/JeffersonLab/coatjava/commits/7ca42a04d89135c4b58af027e13fb658162c7a22/reconstruction/recoil/pom.xml

baltzell (approved):

PR #638, opened 2025-05-14
CLARA I/O service development

baltzell:
* add support for CCDB variation/timestamp
* use RCDB by default for torus/solenoid scales
* add support for sub-schema writing
* add post-processing in the writer
* remove unused EVIO dictionaries
* change OptionParser to require an exact match, no abbreviations
* add default OptionParser option for logging verbosity
* cleanup various log printouts
* reduce logging severity of the most common printouts to FINE
* change default logging to FINE
* add logging flexibility
* import frames i/o services from grapes

Example performance on 32 cores, an RG-D production D2 run, conventional track-finding and no denoising:
Screenshot 2025-05-14 at 19 45 02


build-coatjava.sh
```diff
@@ -60,6 +63,7 @@ done
 
 src_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" &> /dev/null && pwd)"
 prefix_dir=$src_dir/coatjava
+clara_home=$(dirname $prefix_dir)/clara-home
```
c-dilks: why a relative path and not an absolute path?
```suggestion
clara_home=$prefix_dir/clara-home
```

build-coatjava.sh
```diff
@@ -174,6 +178,8 @@ if [ $buildDocs == "yes" ]; then
 fi
 
 # install clara
-#rm -rf clara-home && ./install-clara -c ./coatjava ./clara-home
+if [ $clara == "yes" ]; then
+    rm -rf $clara_home && ./install-clara -c $prefix_dir $clara_home
```
c-dilks: DRY violation: `$clara_home` was already removed above, choose one or the other.

bin/run-clara
```diff
@@ -73,7 +74,9 @@ done
 # Set some JVM options:
 export JAVA_OPTS="$JAVA_OPTS -XX:+IgnoreUnrecognizedVMOptions"
 export JAVA_OPTS="$JAVA_OPTS -Djava.io.tmpdir=$CLARA_USER_DATA -Dorg.sqlite.tmpdir=$CLARA_USER_DATA"
-export JAVA_OPTS="$JAVA_OPTS -Djava.util.logging.config.file=$CLAS12DIR/etc/logging/debug.properties"
+# Set verbosity:
+[ -z ${quiet+x} ] && stub=debug || stub=info
+export JAVA_OPTS="$JAVA_OPTS -Djava.util.logging.config.file=$CLAS12DIR/etc/logging/$stub.properties"
```
c-dilks: `debug.properties` no longer exists; either add that file or change the possible values of `$stub`.

build-coatjava.sh
baltzell: wouldn't that put it inside coatjava?

build-coatjava.sh
baltzell: it's a local variable, want a different name?

baltzell:
let's just remove the clara option, very minor convenience

build-coatjava.sh
c-dilks: yes... this confused me; how about this for clarity:
```bash
clara_home=$src_dir/clara-home
```

c-dilks (commented):
High-level software stuff looks good to me. I'll leave the approval decision to @raffaelladevita (if I click 'approve', the PR will auto merge)

etc/services/evio2hipo.yaml
```diff
@@ -6,11 +6,15 @@ io-services:
...
   - class: org.jlab.service.eb.EBTBEngine
     name: EBTB
 configuration:
-  services:
-    EBTB:
-      dropBanks: 'true'
+  io-services:
+    reader:
+      variation: default
+      timestamp: 05/11/2025-00:00:00
```
raffaelladevita, line 16: Should timestamp and schema_dir be commented out?

raffaelladevita (approved):

etc/services/evio2hipo.yaml
baltzell: Not sure, but it serves as an example of setting a timestamp during decoding

PR #667, opened 2025-06-02
New AI model for AHDC after the geometry change.

mathieuouillon:
This is a new model for the AHDC for the track finding after the modification made to the geometry.


c-dilks:
Ignore the CI job failure of `deploy_web_pages`. It happened because your head branch is also named `development` and that job's conditional does not check the fork name.

c-dilks:
> Ignore the CI job failure of `deploy_web_pages`. It happened because your head branch is also named `development` and that job's conditional does not check the fork name.

But you'd only see this on your fork, not here in this PR

baltzell (approved):

PR #669, opened 2025-06-03
fix: make `install-clara` independent from `libexec`

### LINKED ISSUE #664: install-clara is no longer standalone
due to the automated coatjava version detection via `libexec/`

c-dilks:
Now `install-clara` requires `-c` to specify `coatjava` version, or it'll try to auto-detect.

### DIFF
--- a/install-clara
+++ b/install-clara
@@ -5,7 +5,6 @@ set -e
 # Default versions:
 grapes=2.20
 clara=5.0.2
-coatjava=''
 
 function error() {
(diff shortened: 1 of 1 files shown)


baltzell:
or should we do a pre-commit hook thingy instead?

c-dilks:
> or should we do a pre-commit hook thingy instead?

that's possible, but may cause more trouble than it's worth

baltzell (approved):

PR #672, opened 2025-06-06
Somethings change with pom.xml that make DJL library not loaded.

mathieuouillon:


baltzell:
Can you add this small change to enable AI in the AHDC "unit" test:  https://github.com/JeffersonLab/coatjava/pull/674/files  (Apparently forks are one-way when it comes to PRs)

baltzell (dismissed):

mathieuouillon:
Ok I have added you commit to my pull request

baltzell:
Looks like this might be the maven-enforcer culprit:  
[f9bf333e888](https://github.com/JeffersonLab/coatjava/pull/672/commits/f9bf333e88895c806dc36a8f76aa97afee27a5ee)

parent/pom.xml
```diff
@@ -174,10 +174,69 @@
...
+    </dependency>
+
+    <dependency>
+        <groupId>ai.djl.pytorch</groupId>
+        <artifactId>pytorch-jni</artifactId>
+        <version>2.5.1-0.32.0</version>
+        <scope>runtime</scope>
+    </dependency>
+
+
```
c-dilks: If these aren't shared by other modules, I would just put them all in `reconstruction/alert/pom.xml`

This `dependencyManagement` section is more for dependencies which are needed by many modules.

reconstruction/alert/pom.xml
```diff
@@ -87,16 +87,88 @@
...
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-dependency-plugin</artifactId>
+            <version>3.6.1</version>
+            <configuration>
+                <ignoredUnusedDeclaredDependencies>
+                    <ignoredUnusedDeclaredDependency>ai.djl:model-zoo</ignoredUnusedDeclaredDependency>
+                    <ignoredUnusedDeclaredDependency>ai.djl.pytorch:pytorch-model-zoo</ignoredUnusedDeclaredDependency>
+                    <ignoredUnusedDeclaredDependency>ai.djl.pytorch:pytorch-engine</ignoredUnusedDeclaredDependency>
+                    <ignoredUnusedDeclaredDependency>ai.djl.pytorch:pytorch-native-cpu</ignoredUnusedDeclaredDependency>
+                    <ignoredUnusedDeclaredDependency>ai.djl.pytorch:pytorch-jni</ignoredUnusedDeclaredDependency>
```
c-dilks: Instead of ignoring, I think it would be better to have tests that use these dependencies; otherwise we should probably tell @dependabot to freeze these versions.

parent/pom.xml
mathieuouillon: I have tried that, but then the model is not loaded and I have an error. 
I have never found how to fix it.

reconstruction/alert/pom.xml
c-dilks: These are actually being used in `build-coatjava.sh --unittests`, so it's fine. Ignoring them is the _only_ way to appease `mvn dependency:analyze`, since even scoping the dependencies as `runtime` is not good enough.

c-dilks (approved):

c-dilks:
CI blocked by `clasweb` connection failures.

PR #675, opened 2025-06-06
silence after a few errors

baltzell:


reconstruction/raster/src/main/java/org/jlab/service/raster/RasterEngine.java
```diff
@@ -22,10 +22,11 @@
 
 public class RasterEngine extends ReconstructionEngine {
 
+    private int nErrors = 0;
```
c-dilks: Naive question: is this thread-safe? Do we need it to be?

raffaelladevita: I would think it doesn't matter because:
- the warning is rare and unlikely to occur simultaneously on different threads
- the purpose is just to limit the verbosity

c-dilks: I agree, this case seems safe. We should consider adding data race detection tests though.

c-dilks (approved):

raffaelladevita (approved):

baltzell:
Ran all tests locally, clasweb inaccessible from github all day ...

PR #680, opened 2025-06-10
cleanup docs

baltzell:


c-dilks:
probably should fix the links too.. some of them still point to `clas12-offline-software`.

c-dilks (approved):

PR #688, opened 2025-06-16
build: deploy reconstruction JARs and unify aggregator and parent POMs

c-dilks:
## deploy reconstruction JARs

close #678 

Current deployment only deploys `coat-libs`. New deployment deploys all the `reconstruction` JARs.

<details>
<summary>Click here to show the new deployment tree, for v15.0.0 ...</summary>

```
myLocalMvnRepo/
└── org
    └── jlab
        ├── clas12
        │   ├── detector
        │   │   ├── clas12detector-alert
        │   │   │   ├── 15.0.0
        │   │   │   │   ├── clas12detector-alert-15.0.0.jar
        │   │   │   │   ├── clas12detector-alert-15.0.0.jar.md5
        │   │   │   │   ├── clas12detector-alert-15.0.0.jar.sha1
        │   │   │   │   ├── clas12detector-alert-15.0.0.pom
        │   │   │   │   ├── clas12detector-alert-15.0.0.pom.md5
        │   │   │   │   └── clas12detector-alert-15.0.0.pom.sha1
        │   │   │   ├── maven-metadata.xml
        │   │   │   ├── maven-metadata.xml.md5
        │   │   │   └── maven-metadata.xml.sha1
        │   │   ├── clas12detector-band
        │   │   │   ├── 15.0.0
        │   │   │   │   ├── clas12detector-band-15.0.0.jar
        │   │   │   │   ├── clas12detector-band-15.0.0.jar.md5
        │   │   │   │   ├── clas12detector-band-15.0.0.jar.sha1
        │   │   │   │   ├── clas12detector-band-15.0.0.pom
        │   │   │   │   ├── clas12detector-band-15.0.0.pom.md5
        │   │   │   │   └── clas12detector-band-15.0.0.pom.sha1
        │   │   │   ├── maven-metadata.xml
        │   │   │   ├── maven-metadata.xml.md5
        │   │   │   └── maven-metadata.xml.sha1
        │   │   ├── clas12detector-bg
        │   │   │   ├── 15.0.0
        │   │   │   │   ├── clas12detector-bg-15.0.0.jar
        │   │   │   │   ├── clas12detector-bg-15.0.0.jar.md5
        │   │   │   │   ├── clas12detector-bg-15.0.0.jar.sha1
        │   │   │   │   ├── clas12detector-bg-15.0.0.pom
        │   │   │   │   ├── clas12detector-bg-15.0.0.pom.md5
        │   │   │   │   └── clas12detector-bg-15.0.0.pom.sha1
[...]

PR #690, opened 2025-06-16
fix(build): do not build `parent/pom.xml` as a module

c-dilks:


c-dilks:
nevermind, I'll just do this as part of #688

PR #691, opened 2025-06-16
Use non-default name

baltzell:
in preparation for moving to gitlab


c-dilks:
LGTM. If RG-L needs this, they can change their gitlab repo settings to use this file in their pipelines, rather than the default `.gitlab-ci.yml`.

c-dilks (approved):

PR #696, opened 2025-06-18
test: increase Maven timeout

c-dilks:


c-dilks:
Interesting, in the past ~24 hours Github runners have gone from [receiving timeout responses](https://github.com/JeffersonLab/coatjava/actions/runs/15716274380/job/44286668861) to the much worse [connection refused responses](https://github.com/JeffersonLab/coatjava/actions/runs/15737220204/job/44353080393).

c-dilks:
In other words, this PR does _not_ fix anything, but may help for those with slow access to our Maven repos.

c-dilks:
Default timeout should be long enough; if longer timeouts are desired, these options can be set in `$MAVEN_OPTS`.

PR #706, opened 2025-06-25
Prefer curl

baltzell:
(after JLab's latest filter is killing the wgets)


c-dilks (approved):
Sounds good. But now we get 403'd from Maven...

PR #711, opened 2025-06-27
Minimal github->gitlab CI conversion 

baltzell:
Tested on code.jlab.org (with push-mirroring back to github.com).

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,0 +1,92 @@
+workflow:
+  rules:
+    - if: $CI_PIPELINE_SOURCE == "merge_request_event"
+      auto_cancel:
+        on_new_commit: interruptible
+    - if: $CI_COMMIT_BRANCH == "main"
+      auto_cancel:
+        on_new_commit: conservative
+    - if: $CI_COMMIT_TAG
+      auto_cancel:
+        on_new_commit: conservative
+
+default:
+  interruptible: true
+  before_script:
+    # FIXME: build minimal el9 container, or use container-forge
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -0,0 +1,92 @@
...
+stages:
+  - build-stage
+  - test-stage
+  - deploy-stage
+
+build:
+  stage: build-stage
+  script:
+    - ./build-coatjava.sh -T4 --unittests --spotbugs --quiet --no-progress
+    - tar -czf coatjava.tar.gz coatjava
```
c-dilks: You can just exclude this since the artifact `coatjava/` should preserve permissions (unlike GitHub, where we had to `tar` it).
```suggestion
```

.gitlab-ci.yml
```diff
@@ -0,0 +1,92 @@
...
+      - ARG: electronproton 
+      - ARG: electronprotonC 
+      - ARG: electrongamma 
+      - ARG: electronneutronC
+      - ARG: electronFTpion
+
+jacoco:
+  allow_failure: true
+  stage: test-stage
+  needs: [build]
```
c-dilks, line 58: YAGNI (at least not yet):
```suggestion
```
You could also probably remove the `dependencies:` in most jobs too, since if you don't have `dependencies:`, all artifacts from the previous stage are supposed to be available in the current stage. Up to you.

baltzell: Actually, this one it really explicitly wanted, even though it seemed to me too that it shouldn't .  We can fix it after the move to gitlab.

c-dilks (approved):

PR #736, opened 2025-07-10
Test for github OS changes

baltzell:


c-dilks:
is this testing the cvmfs CI failure? Ubuntu v24 is _supposed_ to be working, they test it upstream. This may be an issue in the action itself

c-dilks:
The first workflow passed, but there was a cache miss (for the cvmfs filesystem); I pushed an empty commit to trigger a new run, to see if anything changes when there's a cache hit.

c-dilks:
Cache hit also works. So rolling back the runner's OS version fixes this, but is a kluge. Let me try flushing the CVMFS cache for Ubuntu 24; hopefully that fixes things, otherwise I'll open an issue in `github-action-cvmfs`.

c-dilks:
Yup... just needed a cache flush. I think we should be good now, so I'm gonna close this.

https://github.com/JeffersonLab/coatjava/actions/runs/16198428834

PR #745, opened 2025-07-16
fix: change hardcoded values

### LINKED ISSUE #744: Change the way columns are displayed for indexed tables
The lengths for the number of indices is hard coded

N-Plx:


N-Plx:
Column access was only working when the index is 3 numbers, made it more general

zieglerv (approved):

PR #754, opened 2025-07-17
fix: upload the `reconstruction` JARs to Maven repo

c-dilks:
I forgot the `scp` command in https://github.com/JeffersonLab/coatjava/pull/688

We could probably just `scp` the whole local repo (`myLocalMvnRepo/`), but I'd rather not accidentally clobber something unrelated; I prefer targeted `scp` commands for safety (but also we have snapshots anyway).


c-dilks:
> I forgot the `scp` command in #688

Therefore 13.0.3 did not deploy the `reconstruction` JARs.

baltzell (approved):

PR #763, opened 2025-07-28
Add fine time correction in ahdc reconstruction

ftouchte:
- Used timestamp et startTime
- Used wf type instead of raw hit cuts


common-tools/clas-detector/src/main/java/org/jlab/detector/pulse/ModeAHDC.java
baltzell, line 1: Those string manipulations should be replaced with bit arithmetic.

ftouchte: very good idea! :+1:

ftouchte:
updated

baltzell (approved):

PR #756, opened 2025-07-17
feat: indexed table class can handle more than 4 indices

N-Plx:
It can handle up to 4 indices with 16 bits or up to 9 indices with 7 bits. Could be made more generic.

PR #768, opened 2025-07-31
update branch

N-Plx:


c-dilks:
FYI, you could have just clicked "Update Branch" at the bottom of #756, instead of making this "backmerge" PR. Less mouse clicks :)

PR #769, opened 2025-07-31
Try a speed test for IndexedTable

baltzell:
good for #756


baltzell:
Good enough for now, from the unit test logs:
```
********************************************************************************************************************
*     BENCHMARK  RESULTS 
********************************************************************************************************************
*   TIMER (IT:GIV      ) : N Calls         9999,  Total Time  =         7.09 sec,  Unit Time =        0.710 msec   *
********************************************************************************************************************
```

c-dilks (approved):
Someday we should generalize these benchmark names, _e.g._, with `this.getClass()`; "IT:GIV" is a bit cryptic for an outsider. If benchmark pretty-printing is a concern, just publish a Markdown table and use the full benchmark name.

PR #784, opened 2025-08-09
Decoder speedups

baltzell:
Adding a dedicated CI test first ...

### DIFF
--- a/common-tools/clas-detector/src/main/java/org/jlab/detector/swaps/SwapTable.java
+++ b/common-tools/clas-detector/src/main/java/org/jlab/detector/swaps/SwapTable.java
@@ -73,17 +73,18 @@ public class SwapTable {
         for (int row=0; row<fromTrans.getRowCount(); row++) {
 
             // crate/slot/channel is the base for mapping between the two tables:
-            final int crate = Integer.valueOf((String)fromTrans.getValueAt(row,0));
-            final int slot = Integer.valueOf((String)fromTrans.getValueAt(row,1));
-            final int channel = Integer.valueOf((String)fromTrans.getValueAt(row,2));
+            final int crate = Integer.parseInt((String)fromTrans.getValueAt(row,0));
+            final int slot = Integer.parseInt((String)fromTrans.getValueAt(row,1));
+            final int channel = Integer.parseInt((String)fromTrans.getValueAt(row,2));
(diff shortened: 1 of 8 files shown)


baltzell:
getting consistently ~40% faster on ifarm/cache, github actions, and lappi

common-tools/clas-detector/src/main/java/org/jlab/detector/swaps/SwapTable.java
```diff
@@ -73,17 +73,18 @@ public SwapTable(IndexedTable fromTrans,IndexedTable toTrans) {
         for (int row=0; row<fromTrans.getRowCount(); row++) {
 
             // crate/slot/channel is the base for mapping between the two tables:
-            final int crate = Integer.valueOf((String)fromTrans.getValueAt(row,0));
-            final int slot = Integer.valueOf((String)fromTrans.getValueAt(row,1));
-            final int channel = Integer.valueOf((String)fromTrans.getValueAt(row,2));
+            final int crate = Integer.parseInt((String)fromTrans.getValueAt(row,0));
```
raffaelladevita: in mon12 we found https://github.com/JeffersonLab/mon12/blob/main/src/main/java/org/clas/detectors/DCmonitor.java#L136-L138 is significantly fasster

raffaelladevita (dismissed):

common-tools/clas-detector/src/main/java/org/jlab/detector/swaps/SwapTable.java
baltzell: than what?

raffaelladevita (dismissed):

baltzell:
See #815, the difference comes from a CCDB conflict for run 809, with LTCC and helicity sharing crate/slot 19/19.

baltzell:
As it turns out, LTCC winning over helicity is correct for 809 (now, with first match in the arbitrarily ordered list winning, instead of last as previously), and the helicity FADC reading in the KPP test were always bogus.  Correct?

baltzell:
And the extra event disappears after only the CCDB changes (CI jobs won't show it until they pickup the weekly CCDB sqlite snapshot).

raffaelladevita (dismissed):

raffaelladevita (approved):

PR #785, opened 2025-08-09
feat:  Add dedicated decoder CI job

baltzell:


c-dilks (approved):
`xrdcp` sometimes may be slow; consider caching the file.

- example from iguana CI: https://github.com/JeffersonLab/iguana/blob/83ced078ba9dbf4a383ef2750345937cc80b47b7/.github/workflows/ci.yml#L32-L54
- then use `actions/cache/restore` in jobs where you need the file

PR #795, opened 2025-08-15
feat:  Auto-detect $CLAS12DIR coatjava location

baltzell:
Currently implemented as a fallback for finding CLAS12DIR from its environment variables or Java system properties, with a System.console check/kludge to make it IDE-compatible.

Need to check whether CLARA hijacks the System.console() ...

### DIFF
--- a/common-tools/clas-utils/src/main/java/org/jlab/utils/system/ClasUtilsFile.java
+++ b/common-tools/clas-utils/src/main/java/org/jlab/utils/system/ClasUtilsFile.java
@@ -8,6 +8,7 @@ import java.io.FileReader;
 import java.io.FileWriter;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.logging.Level;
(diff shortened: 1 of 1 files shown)


common-tools/clas-utils/src/main/java/org/jlab/utils/system/ClasUtilsFile.java
```diff
@@ -37,32 +73,36 @@ public static void printLog(String log){
...
+
         if(envString==null){
             ClasUtilsFile.printLog("Environment variable ["+env+"] is not defined");
             envString = System.getProperty(env);
         }
         
         if(envString == null){
             ClasUtilsFile.printLog("System property ["+env+"] is not defined");
-            return null;
+            if (envString.equals("COATJAVA") || envString.equals("CLAS12DIR")) {
```
c-dilks: `envString` is `null` here, did you mean `env.equals`?
```suggestion
            if (env.equals("COATJAVA") || env.equals("CLAS12DIR")) {
```

c-dilks (approved):

PR #800, opened 2025-08-19
feat:  add option to install coat-libs to local maven repo

baltzell:


baltzell:
unnecessary, already happens w/o this feature, doh ...

PR #801, opened 2025-08-25
Define a default startTime for simulation or ALERT-only reconstruction

ftouchte:
It also fixes another issue : this kind of cuts should be done at the analysis level, not here.

```cpp
if (startTime < 0) { // reject bad events 
 return;
 }
```


mathieuouillon (approved):
Looks good for me. It fixed the issue with the simulation.

PR #805, opened 2025-08-29
Make the reconstruction to work with the modifications in simulation (clas12Tags)

ftouchte:
Feature:

- compute leadingEdgeTime even if the type is bad (the cut on the type can be done at the analysis level)
- if time < 0, set the distance to be 0
- use the time2distance from simulation


baltzell (approved):
Looks like this adds another feature to AHDC's "sim" mode, doing things differently for real/fake data, based on the user's configuration file.

PR #808, opened 2025-09-02
feat: convenience script to run any class's `main`

c-dilks:


c-dilks:
For now I just named it `coatjava`, but better name suggestions are certainly welcome.

baltzell (dismissed):

baltzell:
* coatjava
* coatjavajava
* run-java (don't we have other "run-standard-thing" wrappers?)

c-dilks:
> * run-java (don't we have other "run-standard-thing" wrappers?)

`run-java` would follow our convention (`run-groovy`, `run-jshell`, _etc._) and I think it wouldn't conflict with anything else in the typical `$PATH`. I decided on `run-coatjava` though, since `run-java` sounds a bit too generic...

Unfortunately we're stuck with other generic names, like `run-groovy`; I'd rather them be `run-coatjava-groovy` or something, but downstream Groovy users may depend on `run-groovy`. We could phase out the generic names with name-change warnings.

baltzell (approved):

PR #811, opened 2025-09-03
fix(ci): don't `apt upgrade`

c-dilks:
since it triggers updates for unrelated packages, and sometimes failures such as https://github.com/JeffersonLab/coatjava/actions/runs/17418515756/job/49452235710


c-dilks:
OTOH, `apt update` [is recommended](https://docs.github.com/en/actions/how-tos/manage-runners/github-hosted-runners/customize-runners#installing-software-on-ubuntu-runners)

baltzell (approved):

PR #812, opened 2025-09-03
Use JRE 24 in CI

baltzell:


c-dilks:
this causes the expected LZ4 warning: https://github.com/JeffersonLab/coatjava/actions/runs/17443464249/job/49532409301#step:8:19

PR #815, opened 2025-09-05
Speedups - noclobber/break

baltzell:


baltzell:
@raffaelladevita Seems to indeed restore the event count on kpp test:(

baltzell:
The break causes at least helicity FADCs not to get written ...

baltzell:
yes, /daq/tt/ltcc has crate/slot/channel 19/19/0,2,4 for run 809

PR #819, opened 2025-09-08
feat: added start time reading

### LINKED ISSUE #818: ALERT ATOF bar calibrations
Bar calibrations to be read from CCDB, incomplete bar hits to be kept

N-Plx:


N-Plx:
@ftouchte Hi Felix, I'm adding calibration constant reading to the atof engine. We are loading the tables for the atof and ahdc together from the way you defined the CalibrationConstantLoader, which is completely fine, but then maybe we should move this loading to the ALERTEngine so we don't do it in both the ATOF and AHDC engines. Does that sound right for you? Do you want to make the changes or do you want me to make them? If you make them, please refer to this line to set the proper number of indices for the ATOF tables https://github.com/JeffersonLab/coatjava/blob/cd6cb41ab3ba8230ca2edf44acb6f8345499ba84/reconstruction/alert/src/main/java/org/jlab/service/atof/ATOFEngine.java#L141

ftouchte:
Hello Noémie, it is a good idea. The code is not fresh in my mind, so I think you will be more efficient than me.

- I used to run the reconstruction with the AHDCEngine only. If we move it to the ALERTEngine, that means that I will have to use both of them in this order: ALERTEngine then AHDCEngine. Am I right? Is there a way to directly refer to the ALERTEngine inside the AHDCEngine?

I just have a small modification to propose. [Here](https://github.com/JeffersonLab/coatjava/blob/cd6cb41ab3ba8230ca2edf44acb6f8345499ba84/reconstruction/alert/src/main/java/org/jlab/rec/constants/CalibrationConstantsLoader.java#L1) , I used this package name: `package org.jlab.rec.constants;`. I just notice that it was not very safe. I propose : `package org.jlab.rec.alert.constants;`. (This can be done later.)

baltzell (approved):

PR #825, opened 2025-09-11
Desioned

baltzell:
Based on https://github.com/rtysonCLAS12/C12FD_Denoising and https://jeffersonlab.sharepoint.com/:b:/r/sites/HallBSoftware/Shared%20Documents/General/Denoising_Sep2025.pdf?csf=1&web=1&e=SkGibR


baltzell:
@rtysonCLAS12 @tongtongcao This now works!

tongtongcao:
At https://github.com/JeffersonLab/coatjava/blob/denoising/reconstruction/dn/src/main/java/org/jlab/service/dn/DenoiseEngine.java#L133, new order is set as original order + 10 for noise hits predicted by AI model.
New order seems not to be consistent with order type definition in https://github.com/JeffersonLab/coatjava/blob/development/common-tools/clas-detector/src/main/java/org/jlab/detector/banks/RawBank.java#L66-L79
After merging with background, order of hits could be 0, 10, 20, 30. Then after application of the denoising engine, new order would be plus 10 for noise hits predicted by AI.

PR #826, opened 2025-09-11
Add int-based bank column accessors

baltzell:


common-tools/clas-io/src/main/java/org/jlab/io/hipo/HipoDataBank.java
```diff
@@ -57,7 +57,9 @@ public double getDouble(String path, int index) {
 
     @Override
     public void setDouble(String path, double[] arr) {
-        throw new UnsupportedOperationException("Not supported yet.");
+        int nrows = this.hipoGroup.getRows();
+        for (int i=0; i<nrows; i++)
+            this.hipoGroup.putDouble(path, i, arr[i]);
```
c-dilks, line 62: do we need protection in these `set*` methods for if `i` goes out of the range of `arr`?

baltzell: no, i wouldn't think we want that

raffaelladevita (approved):

PR #770, opened 2025-08-01
feat: QADB binning support

c-dilks:
features:
- new classes to support QADB binning
  - `QadbBin`: a single QADB bin
  - `QadbBinSequence`: a sequence of `QadbBin`s
- both new classes are generic, allowing a `QadbBin` to hold any arbitrary data
- both classes inherit from `DaqScalersSequence` since:
  - `QadbBin` is a sequence of scalers _within_ that bin; the first and last scalers are the bin boundaries
  - `QadbBinSequence` is a sequence of scalers at the bin boundaries
  - Both of these are _subsequences_ of the original `DaqScalersSequence`
  - inheritance from `DaqScalersSequence` is preferred over encapsulation, for cleaner access to protected members without exposing them publicly; furthermore, both `QadbBin` and `QadbBinSequence` are fundamentally sequences of scalers, therefore they are "specializations" of `DaqScalerSequence`

purpose:
- upstream QADB binning implementation out of `clas12-timeline` and make use of `DaqScalersSequence`
- use it in `clas12-timeline` with https://github.com/JeffersonLab/clas12-timeline/pull/367
  - see this PR's description for comparison test results
- this is the first step toward refactoring the "high-level physics" QA timelines to be consistent with the usual "detector" timelines
- now "anything" could be done in QADB bins, if wanted

PR #827, opened 2025-09-11
feat: QADB binning support 

baltzell:
Same as #770, but squashing the last 31 commits like [this](https://stackoverflow.com/questions/5189560/how-do-i-squash-my-last-n-commits-together) to get a commit message like github would've done (?) and then fixing rebase conflicts.


c-dilks:
No need, squash merging will do this, and the PR will retain the messy commit history, if ever needed.

PR #829, opened 2025-09-12
increase ccdb logging level

baltzell:
Includes removing a logging initialization in the DC geometry service (which was blasting away the user's choice).


baltzell:
DC geometry was overriding the logging level ... fixed.

c-dilks (approved):

PR #844, opened 2025-09-22
feat: `SplitLogger`, for sending errors to `stderr` and everything else to `stdout`

### LINKED ISSUE #72: `Logger` level `INFO` should stream to `stdout`, not `stderr`
### Test:
```groovy
// test_reader.groovy

import org.jlab.io.hipo.HipoDataSource

def inHipo = args[0]

def reader = new HipoDataSource()
reader.open(inHipo)
reader.close()
```
Run with:
```bash
run-groovy test_reader.groovy example_hipo_file.hipo >log.stdout 2>log.stderr
```

### Expected result:
`log.stderr` should be empty, indicating no errors (nothing streamed to `stderr`)

### Actual result:
`log.stderr` contains
```
[DataSourceDump] --> opened file with events #
```
which appears to originate from
https://github.com/JeffersonLab/coatjava/blob/40e293da8e9421e1d301166daa5900f3cb213c58/common-tools/clas-io/src/main/java/org/jlab/io/hipo/HipoDataSource.java#L62
This is an `INFO` level printout, and it would be preferable not to have it streamed to `stderr` (especially for the case of running batch jobs, where parsing multiple error logs is essential).

### Feature request:
- send `INFO` to `stdout` and send `WARNING` and `SEVERE` to `stderr`; this may not be easy but something like <https://stackoverflow.com/a/23717493> may be related.
- aside: there are some cases of `System.out.println("ERROR: ...");` which should either be changed to `System.err.println` or use an appropriate `Logger` level

c-dilks:
fixes #72

This PR adds yet another logger `SplitLogger`, which has custom `java.util.logging.Handler`s to send messages with a log level `Level`$\in${`SEVERE`,`WARNING`} to `stderr`, and lower levels to `stdout`.

### Fixes #72 

No more "noise" in error log files (from farm batch jobs); only errors and warnings should appear in error logs, while everything else should be shoved in output (`stdout`) logs.

This "noise" is particularly annoying when `INFO` level messages contain the word "error".

If `stdout` and `stderr` are preferred to be together, so you can see _when_ an error happened in context of other log messages, run interactively; use `stdbuf -o0` in front of your command, to be _sure_ the ordering is correct.

### Formatting

Not really a new thing, but log messages from `SplitLogger` include a prefix:
```
[className] LEVEL: message
```
This is useful to see where log messages are coming from.

It also tells us which classes are using `SplitLogger`, since for the sake of testing this in production, I've only enabled it for classes used by `rebuild-scalers` and `postprocess`; if we like it, we can apply this to all classes and drop `DefaultLogger`, otherwise if we hate it, we can revert this PR.

### Log Level Control

Also not particularly new, but, to control the log level of a `SplitLogger`, make a logging `.properties` file, and set the level of each of your classes for which you want to use a non-default level; use the property name `className.level`, for example:
```
# potato.properties
org.jlab.detector.helicity.HelicityGenerator.level = FINEST
org.jlab.detector.calib.utils.RCDBProvider.level = FINER
```
Then run as
```bash
java -Djava.util.logging.config.file=potato.properties ...
```
or using `run-coatjava`,
```bash
run-coatjava [CLASS_NAME] [CLASS_ARGS] -- -Djava.util.logging.config.file=potato.properties
```
Note that these settings may compete with the log level set by `OptionParser`'s option `-l`; my test on `rebuild-scalers` indicates that `potato.properties` will _override_ what is set by `-l`, which is a nice convenience.

### CLI log level control
I cannot convince the usage of CLI arguments such as `-Djava.util.logging.Level=FINEST` or `-Dorg.jlab.logging.TestSplitLogger.level=FINEST` to be respected; so it seems we need to use a logging `.properties` file (or `OptionParser`'s `-l` option).

### minor edge case
we are no longer getting stack traces from simple `System.exit(101)` calls; if using `DefaultLogger.debug()`, which reads `fine.properties`, and you _just_ call `System.exit(101)` (and not throw an exception), you'll get a nice stack trace like so:
```
java.lang.Throwable: Runtime.exit(101)
	at java.base/java.lang.Shutdown.logRuntimeExit(Shutdown.java:179)
	at java.base/java.lang.Shutdown.exit(Shutdown.java:160)
	at java.base/java.lang.Runtime.exit(Runtime.java:177)
	at java.base/java.lang.System.exit(System.java:1518)
	at org.jlab.analysis.postprocess.Tag1ToEvent.main(Tag1ToEvent.java:45)
```
So it tells you _where_ the `System.exit(101)` call happened. The minimum `.properties` lines that you need are:
```
.level = FINE
handlers = java.util.logging.ConsoleHandler
java.util.logging.ConsoleHandler.level = ALL
```
Using `SplitLogger` stops this nice feature, and the property `java.lang.Runtime.exit.tracing` value does not seem to help; however, I don't think this is a deal-breaker, since instead of just calling `System.exit(101)` silently, we should be sure to always print proper error messages and/or throw runtime exceptions.

Note that general exception stack traces are still printed; you can also use `Logger` to print them at your preferred level, for example, with `SplitLogger` this will send the stack trace to `stderr`:
```java
try {
  throwRuntimeException();
} catch (Exception ex) {
  LOGGER.log(Level.SEVERE, "exception message!", ex);
}
```
Bonus: use this with a lower log level, such as `Level.INFO`, if for some reason you'd rather the stack trace go to `stdout`


baltzell:
Performance wise, seems just one additional check on the integer log level to decide whether to go to stderr or stdout.  Are we pretty sure it's nothing more?  I was thinking like the `logger.debug("Result:  {}", () -> calculate());`, that still works.

baltzell (approved):

c-dilks:
> Performance wise, seems just one additional check on the integer log level to decide whether to go to stderr or stdout. Are we pretty sure it's nothing more? I was thinking like the `logger.debug("Result: {}", () -> calculate());`, that still works.

Yes, one more check for redirection. However, this uses `java.util.logging`, which is apparently not particularly performant anyway, relative to alternatives.

I will try `log4j2`, since it should do everything I'm trying to do here, but faster. It also has async logging, which I think could be good for our multi-threaded use cases.

I'm going to close this PR, but I will make a new branch based on this one to test `log4j2`.

c-dilks:
Nevermind, let's roll with this one, but we should benchmark it before merging.

I tried `log4j2`, but I did not like the long startup time, where it is apparently scanning the large classpath; no way around that, it seems. I also tried `logback`, but was unsuccessful getting the configuration to work. Both were very difficult to set up.

I doubt logging is a bottleneck in our reconstruction.

c-dilks:
> we should benchmark it before merging.

One way to do this would be running `rebuild-scalers` or `postprocess` on a large HIPO file, with the log level set to `FINEST`; compare the time before and after this PR.

baltzell (dismissed):

c-dilks:
@baltzell after your previous approval, I rebased and added a lazy-evaluation test example in https://github.com/JeffersonLab/coatjava/pull/844/commits/cdad7d5aa1016e71fc42db19b35b509860391e4d

baltzell:
I think full classpath is probably too noisy?

c-dilks:
> I think full classpath is probably too noisy?

Yeah, I thought that too. I originally wrote a custom formatter, because I did not like the long class names, but I was afraid that my formatter was slower than `SimpleFormatter`. Unfortunately the short class name is not possible with `SimpleFormatter`, regardless of what name you give the underlying `Logger` instance (which is conventionally the class name in this repo).

raffaelladevita:
It is indeed quite noisy...
In general, I would agree that having the class and method name can be useful, but in our specific case, the class whose name gets printed is typically some common-tools class like the ConstantsManager which is unlikely to be the source of the problem when an error occurs.
Can the printout of the class name be an option?

c-dilks:
> Can the printout of the class name be an option?

Yes, that's a great idea. We could overload `Splitlogger.create`:
```java
  public static Logger create(String name) { return create(name, true); }
  public static Logger create(String name, boolean includePrefix) { ... }
```
Then use `(name, false)` for instances we want to be quiet.

I may also have a way to shorten the class name too, let me see what I can do.

raffaelladevita:
sorry..an accidental clicking

c-dilks:
Now we have
```java
Logger l = SplitLogger.create(ExampleClass.class.getName());
l.info("message");
```
```
[ExampleClass] INFO: message
```
and the quieter
```java
Logger l = SplitLogger.create(ExampleClass.class.getName(), false);
l.info("message");
```
```
INFO: message
```
Do you want the quieter version to also suppress the word `INFO: `?

I was also able to make the class name shorter, since we tend to name our `Logger`s after the class name I could just use the `Logger` name, removing all characters before a dot; we lose the method name though.

baltzell (approved):

PR #848, opened 2025-09-29
feat: add a flag to be read from yaml to choose if the start time is …

### LINKED ISSUE #847: ATOF start time option
Make start time optional for ATOF reading. For now, add a flag to the yaml. Eventually this will need to be all done at event builder level, either with the main EB for clas12 or in a mini EB for alert (alert engine?).

N-Plx:
…used or ignored


N-Plx:
@mathieuouillon This is a fix to avoid requiring the start time for hit reading, with a flag UseStartTime that can be "true" or "false" (quotation marks are important) in the yaml, if you want to test it on simulations that'd be great!

mathieuouillon:
The fix works, thanks. 
Here is an event from simulation using gemc/dev
image

The YAML flag works fine also.

mathieuouillon (approved):
Look good for me. Work with simulation.

PR #849, opened 2025-09-30
use a new waveform classification

ftouchte:
Provide a new waveform classification. 

Complemented by some additional cuts at the reconstruction level, we have seen an increase of the detection efficiency. 

One can refer to the study presented during the ALERT meeting on September 29, 2025: https://clasweb.jlab.org/wiki/images/b/b0/WfType-study.pdf


mathieuouillon (approved):
Look good for me. The decoding and reconstruction work. 

image
image
image

baltzell:
Note, this adds potential thread-safety issues.  Looks like ModeAHDC's extract already wasn't thread safe.  We'll want to fix that before using this in reconstruction (meanwhile, single-threaded usage in the decoder is ok).

c-dilks:
> Note, this adds potential thread-safety issues. Looks like ModeAHDC's extract already wasn't thread safe. We'll want to fix that before using this in reconstruction.

opened #850, so we can start testing for these types of issues

baltzell (approved):

PR #853, opened 2025-10-01
feat:  add new DC denoising

baltzell:


baltzell:
The current order marking in denoising is defined [here](https://code.jlab.org/hallb/clas12/coatjava/denoising/-/blob/main/denoising/code/drift.cc?ref_type=heads#L162-198).

baltzell:
Need to add the order marking fixes and latest network from @tongtongcao and @rtysonCLAS12.

raffaelladevita (approved):

PR #867, opened 2025-10-08
use FastMath in t2d table loading

baltzell:


tongtongcao:
FastMath is not accurate. For example, FastMath.cos(Math.toRadians(60.)) = 0.50022143 instead of 0.50000000
I found this issue long time ago, so changed FastMath into Math for everywhere in DC reconstruction.

PR #869, opened 2025-10-09
fix: suppress restricted method warnings

### LINKED ISSUE #866: Unnamed module, native access warnings
Maybe new in 25:
```
WARNING: A restricted method in java.lang.System has been called
WARNING: java.lang.System::loadLibrary has been called by net.jpountz.util.Native in an unnamed module (file:/ceph24/hallb/ceph/testing/sw/coatjava/clara/plugins/clas12/lib/clas/coat-libs-13.3.0-SNAPSHOT.jar)
WARNING: Use --enable-native-access=ALL-UNNAMED to avoid a warning for callers in this module
WARNING: Restricted methods will be blocked in a future release unless native access is enabled
```
and the same for `org.fusesource.jansi.internal.JansiLoader`.

c-dilks:
fix #866

### DIFF
--- a/libexec/env.sh
+++ b/libexec/env.sh
@@ -32,6 +32,10 @@ if [ "${1-}" != '--no-classpath' ]; then
   export COATJAVA_CLASSPATH
 fi
 
+# java options
+# suppressions of "WARNING: A restricted method in java.lang.System has been called"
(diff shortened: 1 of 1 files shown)


c-dilks:
Not working, abandoning this for now...

PR #874, opened 2025-10-10
feat:  support "stock" schema in YAML files

baltzell:


common-tools/clas-io/src/main/java/org/jlab/utils/ClaraYaml.java
```diff
@@ -93,6 +94,44 @@ public String schemaDirectory() {
...
+   
+    /**
+     * Set the schema_dir, assuming it's currently the short (basename) of a
+     * "stock" schema, but only if it doesn't already look like an absolute path.
+     */
+    public final void setStockSchemaDirectory() {
+        String s = getSchemaDirectory();
+        if (s != null && !s.trim().startsWith("/")) 
+            setSchemaDirectory(getStockSchemaDirectory(s));
+    }
```
c-dilks, line 124: should we handle the case when `getSchemaDirectory()` returns `null`?

baltzell: null means "doesn't exist" here, probably the method needs a better name, but the behavior is intended

c-dilks (approved):

baltzell:
other option is a new YAML key

PR #879, opened 2025-10-10
saving FMT tracks in REC::FTrack

raffaelladevita:


reconstruction/eb/src/main/java/org/jlab/service/eb/EBEngine.java
```diff
@@ -221,6 +225,11 @@ public boolean processDataEvent(DataEvent de,EBScalers ebs) {
                     DataBank x = DetectorData.getUTracksBank(cutracks, ctracks, de, utrackBank);
                     de.appendBanks(x);
                 }
+
+                if (!ftracks.isEmpty()) {
+                    DataBank x = DetectorData.getUTracksBank(ftracks, tracks, de, ftrackBank);
```
baltzell: getUTracksBank?

baltzell (approved):

PR #880, opened 2025-10-12
fix: increase heap max for `rebuild-scalers`

c-dilks:
RG-A runs 4282 and 4293 require a bit more heap space for `org.jlab.analysis.postprocess.RebuildScalers`

### DIFF
--- a/bin/rebuild-scalers
+++ b/bin/rebuild-scalers
@@ -2,9 +2,11 @@
 
 . `dirname $0`/../libexec/env.sh
 
+split_cli $@
+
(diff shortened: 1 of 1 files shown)


baltzell:
does the runtime option we added recently not work?

c-dilks:
> does the runtime option we added recently not work?

yes, let's do that instead

baltzell:
let's just increase the default too

c-dilks:
done

baltzell (approved):

PR #896, opened 2025-10-18
Add uber services

baltzell:
And corresponding `data-*-uber.yml` steering files.


etc/services/data-ai-uber.yml
```diff
@@ -0,0 +1,34 @@
+configuration:
+  global:
```
raffaelladevita, line 2: should probably include these commented lines as in the old yamls
```
#    timestamp: 12/31/2020-00:00:00
## uncomment the following two lines for compatibility with alignments before the DC fixes
#    dcMinistagger: "NOTONREFWIRE"
#    dcFeedthroughs: "OFF"
```

etc/services/data-ai-uber.yml
```diff
@@ -0,0 +1,34 @@
...
+    variation: rgd_fall2023
+  io-services:
+    writer:
+      schema_dir: dst
+  services:
+    1ST:
+      magfieldSolenoidMap: Symm_solenoid_r601_phi1_z1201_13June2018.dat
+      magfieldTorusMap: Full_torus_r251_phi181_z251_25Jan2021.dat
+    HB:
+      run: "6302"
```
raffaelladevita: maybe switch to the RG-D run number

raffaelladevita (changes requested):

raffaelladevita:
Tested comparing output obtained with the uber services and the regular ones. No difference was found in 1000 events

raffaelladevita (approved):

PR #854, opened 2025-10-02
Use JRE 25 in CI jobs

baltzell:

PR #918, opened 2025-10-28
fix: use consistent JDK version for CI and POM

c-dilks:
- reverts PR #854, since we should test with the release specified in the POM file, and what is currently in the latest versioned `clas12` environment module. If we test with a later version, we risk a dependency requiring a later JDK, and it won't be noticed by the CI.
- JDK 25 will still be tested, but restricted to unit tests only
- introduces string `JAVA_VERSION`, to indicate all the places we need to update when we want to increase the Java version

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,8 +21,10 @@ defaults:
     shell: bash
 
 env:
-  java_version: 25
+  JAVA_VERSION: 21 # must be consistent with POM; see also dependabot config for any version limits
+  JAVA_VERSION_LATEST: 25 # the latest `JAVA_VERSION` that we test with CI
   java_distribution: zulu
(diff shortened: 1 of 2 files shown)


baltzell:
let's do separate jobs for 21 and 25

c-dilks:
> let's do separate jobs for 21 and 25

That could be done in another PR; this PR is meant as a `fix`. Some approaches:
- expand the matrix jobs
- use a reusable workflow, with the Java version as an input variable
- manual workflow, with the Java version as an input variable
- create a draft PR proposing Java 25 in the CI, for occasional manual testing by rebasing

c-dilks:
The unit test job is now a job matrix, for testing JDK versions `JAVA_VERSION=21` and `JAVA_VERSION_LATEST=25`.

baltzell (approved):

PR #935, opened 2025-11-04
use the same time2distance for simu and real data

ftouchte:
We now invert the ccdb time2distance in simulation.

https://github.com/gemc/clas12Tags/pull/58


baltzell (approved):

baltzell:
We'd like to have a new COATJAVA release this week.   Would it be best to wait to merge this until the corresponding changes are released in GEMC/5.12?  I think that is next week.  On the other hand, if this isn't really breaking much, e.g., if there's already significant GEMC/COATJAVA inconsistencies for AHDC, maybe this PR can/should go ahead and be merged?

ftouchte:
This PR can wait the next GEMC release as it has no effect on real data.

ftouchte:
A new release of GEMC is available. I think we can push this commit.

https://github.com/gemc/clas12Tags/releases/tag/5.12

baltzell:
It doesn't appear to be fully available yet, but I guess we should go ahead?

baltzell (approved):

PR #949, opened 2025-11-15
Decodering

baltzell:


baltzell:
Now less than 3 ms per event, on ifarm and my laptop.

baltzell:
superceded by #979

PR #950, opened 2025-11-15
fix:  spotbugs addon

baltzell:
Removed unused/unnecessary singletons.  Restore non-synchronized singleton access to MagneticFields.  Use eager singleton initialization where appropriate.


baltzell:
These are always initialized single-threaded and probably shouldn't be synchronized:
* DCdatabase
* Magfield
*

PR #958, opened 2025-11-21
Add magfield and test data as git submdules

baltzell:


.github/workflows/ci.yml
```diff
@@ -87,6 +87,8 @@ jobs:
         run: libexec/version-bump.sh ${{ github.ref_name }}
       - name: build
         run: |
+          git lfs install
+          git submodule update --init ./etc/data/magfield
           ./build-coatjava.sh --cvmfs --no-progress -T${{ env.nthreads }}
```
c-dilks, line 92: You could drop the `--cvmfs` now; I guess you could use `--nomaps`, or better yet, add an option `--lfsmaps` to run `git lfs ... && git submodule update ...`

c-dilks:
I think this approach is good enough. I had one comment about `build-coatjava.sh` options above.

PR #964, opened 2025-11-24
feat: `SplitLogManager` for treewide logging control

c-dilks:
## Description
Changes:
- convert `SplitLogger` into `SplitLogManager`
  - `SplitLogger.create` is replaced by `SplitLogManager` methods that override `LogManager`, such that `Logger.getLogger` and similar commands instantiate a logger that has the same properties as the old `SplitLogger`
  - it is now easy to swap to a different custom log manager

This is a much cleaner approach for controlling our logger; now all we need to do is:

1. Use system property to choose a log manager
   - in this case, the new `SplitLogManager`, via a `java` CLI option:
     ```bash
     -Djava.util.logging.manager=org.jlab.logging.SplitLogManager
     ```
   - all `bin/` scripts now use this option (via `$JAVA_OPTS`, see below)
1. In your Java class, instantiate a `Logger` like normal:
   - just use `Logger.getLogger`, which will use the manager set by the `java.util.logging.manager` property:
     ```java
     static final Logger logger = Logger.getLogger(MyClass.class.getName());
     ```

## Log Levels

- No changes in `coatjava` itself; see the 'Developer Notes', linked from the main `README.md`
- Added log-level control to `run-clara` with new option `-L`
  - default level is `INFO` (`-Linfo`)
  - for example, `-Loff` to silence (mostly) everything, or `-Lall` for _very_ verbose
  - option `-q` for quiet output is (mostly) unchanged

## About `$JAVA_OPTS`
The environment variable `$JAVA_OPTS` is not used by the `java` command; rather it is a convention for things which wrap `java`. I added a default `$JAVA_OPTS` setting in `libexec/env.sh` that:
- sets the log manager
- prioritizes user's `$JAVA_OPTS`, if they have any
- all `java`-calling `bin/*` scripts now use `$JAVA_OPTS`, but as the _lowest_ priority in the CLI option ordering

Therefore, for all `java`-calling `bin/*` scripts, we have the following priority order of `java` options (from lowest priority to highest, _i.e._, left-to-right):
1. `$JAVA_TOOL_OPTIONS` and `_JAVA_OPTIONS`: JVM options; we don't appear to use them here, but they are used by `java` at the _lowest_ priority
1. `$JAVA_OPTS`: includes the log manager choice (set in `env.sh`), followed by user's `$JAVA_OPTS`, if any
1. hard-coded options unique to each `bin/*` script, such as `-Xms` and `-Xmx`
1. some scripts allow for further override options, passed to the script as arguments to the right of `--` (see `env.sh`'s function `split_cli`); these will take the _highest_ priority, since they are meant to be overrides

### DIFF
--- a/bin/bg-merger
+++ b/bin/bg-merger
@@ -6,7 +6,7 @@ export MALLOC_ARENA_MAX=1
 
 split_cli $@
 
-java -Xmx1536m -Xms1024m -XX:+UseSerialGC ${jvm_options[@]} \
+java ${JAVA_OPTS-} -Xmx1536m -Xms1024m -XX:+UseSerialGC ${jvm_options[@]} \
     -cp ${COATJAVA_CLASSPATH:-''} \
     org.jlab.analysis.eventmerger.EventMerger \
     ${class_options[@]}
--- a/bin/bos2hipo
+++ b/bin/bos2hipo
@@ -2,7 +2,7 @@
 
 . `dirname $0`/../libexec/env.sh
 
-java -Xms1024m -XX:+UseSerialGC \
+java ${JAVA_OPTS-} -Xms1024m -XX:+UseSerialGC \
     -cp ${COATJAVA_CLASSPATH:-''} \
     org.jlab.io.utils.Bos2HipoEventBank \
     $*
--- a/bin/cvt-compare
+++ b/bin/cvt-compare
@@ -4,7 +4,7 @@
 
 export MALLOC_ARENA_MAX=1
 
-java -Xmx1536m -Xms1024m -XX:+UseSerialGC \
+java ${JAVA_OPTS-} -Xmx1536m -Xms1024m -XX:+UseSerialGC \
     -cp ${COATJAVA_CLASSPATH:-''} \
     org.jlab.rec.cvt.Geometry \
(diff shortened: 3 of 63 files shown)


bin/bg-merger
```diff
@@ -6,7 +6,7 @@ export MALLOC_ARENA_MAX=1
 
 split_cli $@
 
-java -Xmx1536m -Xms1024m -XX:+UseSerialGC ${jvm_options[@]} \
+java ${JAVA_OPTS-} -Xmx1536m -Xms1024m -XX:+UseSerialGC ${jvm_options[@]} \
```
c-dilks, line 9: Double check there aren't any _new_ `bin/` scripts which also need to be updated.

c-dilks: no new ones yet...

c-dilks:
@baltzell @raffaelladevita at long last this is ready. I confirm it also works well with `clara`.

However, `clara`'s output logs aren't split to `stderr`/`stdout`, so I [propose a fix for that too](https://code.jlab.org/hallb/clas12/coatjava/clara-java/-/merge_requests/4).

baltzell (approved):

PR #966, opened 2025-11-26
test(ci): use LFS for `validation-data`

c-dilks:
- test getting `validation-data` files from LFS
- we haven't tested this yet, since we always have cache hits
- we'll _need_ this to work if the relevant cache files get flushed


c-dilks:
fixed by enabling LFS access in repo settings.

PR #981, opened 2025-12-04
ci: comment details on dependabot submodule bump PRs

c-dilks:
Dependabot's PRs for submodules do not provide any info about what changed; there is no way to configure dependabot to provide more details.

This PR adds an action to automatically comment on dependabot PRs what has changed.

An example of such a comment is below, tested on #980.


c-dilks:
- confirmed works for #980 
- now rebasing onto `development` and restricting trigger to dependabot only

baltzell (approved):

PR #1011, opened 2025-12-16
Prefer pi+ over pi- for start time determination

baltzell:
New priority:
1. e<sup>-</sup>
1. e<sup>+</sup>
1. 𝜋<sup>-</sup>
1. 𝜋<sup>+</sup>

Also need to update documentation.

### DIFF
--- a/reconstruction/eb/README.md
+++ b/reconstruction/eb/README.md
@@ -136,8 +136,8 @@ particle must have both
 1.  a track in the DC
 2.  an associated FTOF hit in panel 1B (preferred) or 1A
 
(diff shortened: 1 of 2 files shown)


c-dilks:
What impact will this have?

baltzell:
The only effect will be for non-electron triggers.  Now (previously) the highest momentum negative (positive) track will be assumed a 𝜋<sup>-</sup> (𝜋<sup>+</sup>) and used to determine the event start time.

raffaelladevita (approved):

PR #1012, opened 2025-12-17
Iss1006 vtx speed3

### LINKED ISSUE #1006: vertexing speed

raffaelladevita:
Reduces the computing time of the overtaxing service by limiting the maximum swimming path length
Tests on RG-E data show a reduction by a factor > 4.


baltzell:
And 0.5 meters is sufficient?  That's relative to the starting point of the swim?

raffaelladevita:
Yes, it is relative to the starting point of a swim. For each pair of tracks with opposite charge, swimming is first used to get the particles’ trajectory in a z range of [-2 : 12] cm around the particle vz. 50 cm should be plenty for this.

zieglerv (approved):

PR #1031, opened 2025-12-23
Cleanup swim wrapper

baltzell:
In preparation for new swimming versions.

### DIFF
--- a/common-tools/swim-tools/src/main/java/org/jlab/clas/swimtools/Swim.java
+++ b/common-tools/swim-tools/src/main/java/org/jlab/clas/swimtools/Swim.java
@@ -1,24 +1,30 @@
 package org.jlab.clas.swimtools;
 
-import cnuphys.adaptiveSwim.AdaptiveSwimException;
+import java.util.ArrayList;
+import java.util.List;
+
+import org.apache.commons.math3.util.FastMath;
+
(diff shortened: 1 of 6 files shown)


common-tools/swim-tools/src/main/java/org/jlab/clas/swimtools/Swim.java
```diff
@@ -66,29 +65,26 @@ public Swim() {
...
-     * @param y0
-     * @param z0
-     * @param thx
-     * @param thy
-     * @param p
+     * @param direction +1 for out -1 for in
+     * @param x0 (cm)
+     * @param y0 (cm)
+     * @param z0 (cm)
+     * @param thx (units?)
```
zieglerv, line 83: Variables thx and thy are dimensionless

common-tools/swim-tools/src/main/java/org/jlab/clas/swimtools/Swim.java
```diff
@@ -66,29 +65,26 @@ public Swim() {
...
+     * @param thy (units?)
+     * @param p (units?)
```
zieglerv, line 85: Variable p unit is GeV

zieglerv (dismissed):

PR #1046, opened 2026-01-06
fix:  add check for null

baltzell:
Happens regularly in RG-L, something like 1 in 10k events.


tongtongcao:
Good finding. It is a safe change. Actually, null check for vertex point takes place in the following function matchHits with  stateVecs as a parameter.

tongtongcao (approved):

PR #1057, opened 2026-01-12
apply AI model for HB tracking

tongtongcao:
A new engine, called as DCHBTrackingAI, is added to apply AI models for estimation of HB tracks.
Then, there are three engines for HB tracking:
1. DCHBPostClusterConv, where all possible cluster combos are tried, and tracking is processed with KF
2. DCHBPostClusterAI, where cluster combos are predicted by AI, and tracking is processed with KF
3. DCHBTrackingAI, where cluster combos are predicted by AI, and track states are estimated by AI

As test, HB tracks by AI has better resolution than HB tracks by KF. 
See details in slides: 
[HB Tracking by AI.pdf](https://github.com/user-attachments/files/24569105/HB.Tracking.by.AI.pdf)

### DIFF
--- a/common-tools/clas-tracking/src/main/java/org/jlab/clas/tracking/kalmanfilter/zReference/KFitter.java
+++ b/common-tools/clas-tracking/src/main/java/org/jlab/clas/tracking/kalmanfilter/zReference/KFitter.java
@@ -31,7 +31,7 @@ public class KFitter extends AKFitter {
     
     private static final double initialCMBlowupFactor = 70;
     
-    public StateVecs sv = new StateVecs();
+    private StateVecs sv = new StateVecs();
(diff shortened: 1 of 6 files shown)


common-tools/clas-tracking/src/main/java/org/jlab/clas/tracking/kalmanfilter/zReference/KFitter.java
baltzell, line 1: you really want to make those variable public?

tongtongcao: Not really. Just for convenience when code development, but forgot to change back.

baltzell (approved):

PR #1059, opened 2026-01-12
fix:  transform DC covariance matrix for TB tracks at vertex from local to global

tongtongcao:
The software update for CM transformation was taken long time ago.
As test, transformed CM is reasonable.
CM is sensitive to DC hit resolution.
After new hit resolution is defined, results for transformation of CM will be re-visited.
See details in slides:
[Transformation of Covariance Matrix.pdf](https://github.com/user-attachments/files/24573097/Transformation.of.Covariance.Matrix.pdf)


baltzell (dismissed):
We should probably just replace the old covariance matrix bank w/ this new schema.

baltzell (approved):

baltzell (dismissed):

baltzell:
needs an approval

baltzell:
Oops, the DST banks also need description updates.

baltzell (approved):

PR #1051, opened 2026-01-07
Decoding speedups

baltzell:
Summary:
* Less string comparisons and looping
* More hash/index-based access for CCDB tables and banks
* Use new global `TranslationTable` class, with additional `DetectorType` field
* Reduce EVIO branch/node transversing

Side Effects:
* Run 5038
  * Row-ordering changed for `ECAL::tdc`, `RAW::vtp`, `BMT::adc`, and `RAW::adc`
  * CCDB overlap for `/daq/fadc/ltcc` and `/daq/fadc/rf` (`LTCC::adc` and `RF::adc`)
* Run 18799
  * CCDB overlap for `/daq/tt/band` and `/daq/tt/raster` (`BAND::adc` and `RASTER::adc`)

_Note, this is the completion of #1008 ..._

PR #1065, opened 2026-01-13
fix: reduce EVIO branch transversing in decoding, cleanup

baltzell:
Note, this changes the row-ordering in many raw HIPO banks, now sorted by crate.  That change can be ignored with  `hipo-diff`'s new sorting option.

_That crate number currently exists only in the online CODA database (and resulting EVIO/HIPO files)._

_Note, this was a component of #1051._


baltzell:
Note, this only changes row ordering.  Any reason not to merge?

raffaelladevita (approved):

PR #1068, opened 2026-01-15
fix:  improve scaling script, fix broken column numbers

baltzell:


baltzell:
Now looks like this:  [https://userweb.jlab.org/~baltzell/scaling.svg](https://userweb.jlab.org/~baltzell/scaling.svg)

PR #1074, opened 2026-01-20
Updating how time offsets are applied, dealing with different start t…

### LINKED ISSUE #1069: Time offsets for ALERT ATOF

N-Plx:
…imes


baltzell:
What's the reason for the `float->Float` switch?

baltzell:
> What's the reason for the `float->Float` switch?

Probably just for the convenience of null, which is just fine if it's not called much.  But for a low-level "hit" class that gets instantiated a lot (?), keep in mind Float incurs more JVM memory garbage.

baltzell (approved):

N-Plx:
Yes the float->Float is to be able to use null. Not sure what is safer memory-wise, I'm open to updating it if you have advice!

baltzell:
It's not about "safety", just memory usage.

PR #1080, opened 2026-01-22
fix: run-clara doesn't work on macos

### LINKED ISSUE #1079: run-clara doesn't work on macos
Probably ~~running under zsh instead of the bash shebang~~ ... `run-clara: line 98: ${java_log_level^^}: bad substitution`

c-dilks:
fixes #1079


baltzell:
just bashisms vs posix(?) stuff?

c-dilks:
> just bashisms vs posix(?) stuff?

Licensing concerns -> macOS ships with 20-year old Bash

baltzell (approved):

PR #1084, opened 2026-01-30
fix RAW::tdc.TDC column index

baltzell:


raffaelladevita (commented):
RAW::adc has the same problem

raffaelladevita (approved):

PR #1089, opened 2026-02-03
Update ahdc geometry

ftouchte:
The current geometry in `coatjava` is very old and slightly differs from the final design of the detector.
- The end points are in z=-188 mm and z=+162.5 mm instead of ±150 mm
- The stereo angles are not exactly 20° but : 19.1489°, 19.2857°, 20.0°, 20.6897°, 20.0° respectively for each super layer.

Corresponding PR in clas12Tags: https://github.com/gemc/clas12Tags/pull/65


baltzell:
We'll have a new release of coatjava, and then gemc, for this.  This week.  The last before moving to real run number simulations.

baltzell (approved):

PR #1100, opened 2026-02-09
Move Kalman Filter in ALERT engine

ftouchte:
- Move Kalman Filter in ALERT engine
  - Require/Read AHDC::track and AHDC::hits form AHDCEngine
  - Remove redundant attributes in the object Track (delete everything with the _kf extension)
  - Initialise the Kalmaan filter with Niter (number of iterations) and particle (PDGParticle type)
  - Update AHDC::hits in ALERT engine after the Kalman Filter


baltzell:
Is this now loading the AHDC geometry twice?  (Once in AHDC's engine and once in ALERT's)

ftouchte:
Yes. 

I noticed that there is something similar for the ATOF. 

I think it is one of the limitation of using three different engines.

baltzell:
It's not a limitation of the 3 engines, but we can address it later if necessary.  It's only memory:)

baltzell:
The approve button seems to have disappeared since last week.  Anyone else seeing the same?

baltzell (approved):

PR #1105, opened 2026-02-12
Add programs to use doca in helixfitting.

zyzhangjlab:
Add programs to use doca to refine positions of clusters, then use them in helixfitting.


baltzell:
I think we need someone more familiar with ALERT tracking to approve this.  @ftouchte @mathieuouillon @efuchey ?

baltzell:
The github CI jobs are fixed and should pass after getting this branch up to date.

ftouchte (approved):

whit2333:
FWIW I don't  think we need any "Clusters". I don't think clustering is a good idea for the tracking hits. What is needed is just a fitter that uses the DOCA for each hit appropriately.

PR #1108, opened 2026-02-13
fix: correct calculated doca in KF tracking

### LINKED ISSUE #1107: Correct calculated doca in KF tracking
In KF tracking, calculated doca is defined as projection point of fit trajectory and wire line, but doca should be the closest approach between trajectory and wire line. Therefore, calculated doca should be corrected.

tongtongcao:
In KF tracking, calculated doca is defined as projection point of fit trajectory and wire line, but doca is supposed to the closest approach between trajectory and wire line. Therefore, calculated doca is corrected by alpha at TB level.
As test, the update fixes issue for bias of fit residual.
No effect on HB tracking.
Change on TB track is slight since effects on L&R hits offset and hit error is scaled by 2 during tracking.

### DIFF
--- a/common-tools/clas-tracking/src/main/java/org/jlab/clas/tracking/kalmanfilter/AMeasVecs.java
+++ b/common-tools/clas-tracking/src/main/java/org/jlab/clas/tracking/kalmanfilter/AMeasVecs.java
@@ -10,6 +10,7 @@ import org.jlab.geom.prim.Arc3D;
 import org.jlab.geom.prim.Line3D;
 import org.jlab.geom.prim.Point3D;
 import org.jlab.geom.prim.Transformation3D;
+import org.jlab.geom.prim.Vector3D;
 
 /**
  *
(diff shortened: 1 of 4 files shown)


raffaelladevita:
Tested successfully on straight track data, see 
[20260211_residuals.pdf](https://github.com/user-attachments/files/25346449/20260211_residuals.pdf)

raffaelladevita:
Tests on production data show this eliminates multimodal shapes in residuals vs. doca in R1 and R3, but makes R2 worse.

Checking further, it looks like this is due to https://github.com/JeffersonLab/coatjava/blob/development/reconstruction/dc/src/main/java/org/jlab/rec/dc/hit/FittedHit.java#L377 that makes alpha different from the track local angle

tongtongcao:
For the update of calculated doca in KF processing, doca is defined as the closest approach between wire line and trajectory line at a given DC layer, where the trajectory line is defined as point (x, y, z) and direction (tx, ty, 1). 
To be consistent with LR of measured doca, sign of calculated doca still depends on x of projection point and wire line.
Correspondingly, matrix H is updated with doca dependent of x, y, tx and ty.

zieglerv (approved):
Code modifications look fine.

PR #1109, opened 2026-02-13
fix cvmfs github CI issues with 5.4

baltzell:


c-dilks:
works, because otherwise we pick up the `v5` tag, not the _latest_ `v5.4` tag

c-dilks (approved):

PR #1120, opened 2026-02-20
Clas12swimtest

heddle:
Added apache jar


c-dilks:
Looks like there are unit test failures. You can run unit tests locally with
```bash
./build-coatjava.sh --unittests
```
Here is the relevant CI log from the last run: https://github.com/JeffersonLab/coatjava/actions/runs/22238531376/job/64336135602#step:5:5016

baltzell:
There's a 1-line change in `common-tools/swim-tools/src/main/java/org/jlab/clas/swimtools/Swim.java`, that looks like a possible culprit for the test failure.

PR #1066, opened 2026-01-13
fix: reduce string comparisons in ADC decoding, cleanup

baltzell:
Note, this assumes unique `crate/slot/channel` in all `/daq/fadc` CCDB tables.  Now, the first match wins.  Previously the last match won.  For run 5038, the `RF/LTCC::adc` banks have a resulting difference, which will be addressed by fixing CCDB tables to be unique.

_Note, this was a component of #1051._

PR #1124, opened 2026-02-21
feat:  add convenience scripts for diffing gold runs

baltzell:
Used for testing PRs #1064, #1065, and #1066.  Likely to be useful again soon.


libexec/gold-run
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+exe=decoder4u -n 10000
+csv=gold.csv
+cache=$(cd $(dirname ${BASH_SOURCE[0]}) &> /dev/null && pwd)/pin-gold.txt
```
c-dilks: These files are not stored in `coatjava`; at least inform the user if their current working directory is wrong. Also, do you mean `gold.pin`?

baltzell: Did we decide where `gold.csv` should live?  (`pin-gold.txt` should live beside it)  I think I was thinking here:  https://code.jlab.org/hallb/clas12/raw-data

baltzell: this should probably do it

c-dilks (approved):

PR #1132, opened 2026-02-27
cleanup rg-l test yaml

baltzell:


etc/services/rgl-clarode.yml
```diff
@@ -57,3 +44,21 @@ services:
     name: EBTB
   - class: org.jlab.rec.rich.RICHEBEngine
     name: RICH
+  - class: org.jlab.service.atof.ATOFEngine
+    name: ATOF
+  - class: org.jlab.service.ahdc.AHDCEngine
+    name: AHDC
+configuration:
+  global:
+    variation: rge_spring2024
```
raffaelladevita: could use the rgl_spring2025 variation that has the. new alignment

baltzell: done

raffaelladevita (approved):

PR #1137, opened 2026-03-05
Fix N_hits issue & fill AHDC::track bank before KF

zyzhangjlab:
1. Fix the issue that there are events with number of hits < 6.
2. fill chi2, sum_adc, path and dEdx in AHDC::track bank before Kalman Filter.


ftouchte:
I made some review. The idea is that, if a quantity can be calculated, we should not have a setter for it. 

Example: `sum_adc` can be directly calculated, I guess the path is set. But `dEdx` uses both of these information. 

We should make a choice. Whether we compute everything, whether we set everything.

As a consequence, some variables are left non used.

baltzell:
@ftouchte et.al, if you think this should be merged, please approve it.

ftouchte (approved):

baltzell:
just need to get the fork branch up to date in order to merge

PR #1138, opened 2026-03-06
Include the charge in the energy loss computation

ftouchte:
Can I add this feature?

image


baltzell:
Can you add some small javadoc to those two methods?  e.g., one should clearly say it assume charge=1.

baltzell (approved):

PR #1165, opened 2026-03-25
correct the calculation of z positions of  clusters and docaclusters

zyzhangjlab:


baltzell:
I think @hauenst mentioned this is related to geometry consistency between simulation and reconstruction?

reconstruction/alert/src/main/java/org/jlab/rec/ahdc/Cluster/Cluster.java
```diff
@@ -26,14 +29,38 @@ public class Cluster {
...
 	private ArrayList<PreCluster> _PreClusters_list;
 
+	private static Line3D representativeLine(PreCluster pc) {
+		if (pc == null || pc.get_hits_list() == null || pc.get_hits_list().isEmpty()) {
+			return null;
+		}
+		Hit h = pc.get_hits_list().get(0);
+		return h.getLine();
+	}
+	public static double getStereoAngleDeg(int superlayerId) {
```
whit2333: Why is this here? This is not using "single source of geometry"  information.

whit2333 (changes requested):

reconstruction/alert/src/main/java/org/jlab/rec/ahdc/Cluster/Cluster.java
zyzhangjlab: I mean to get the origin and end z position from geometry with one line from one precluster.

ftouchte: @zyzhangjlab I think it is about the stereo angles. Can you find a way to retrieve them using only the geometry? We don't want hard coded values.

zyzhangjlab: @ftouchte I see I can calculate them using only the geometry. I will do this later.

ftouchte (dismissed):

reconstruction/alert/src/main/java/org/jlab/rec/ahdc/Cluster/Cluster.java
```diff
@@ -4,16 +4,19 @@
...
 /**
  * Cluster are compose by 2 PreCluster on layer with a different stereo angle
  */
 public class Cluster {
 
 	private int                   _trackId = -1;
-	private double                _StereoAngle = 20.0;
-	private double                _DeltaZ = 300.0;
-	private double                _Zoffset = 150.0;
+	private double                _StereoAngle = 20.0;//not used
```
mathieuouillon: If these variables are not used anywhere, they should be removed (lines 17 to 19)

mathieuouillon (commented):
Look good, just remove the unused variables

baltzell (approved):

baltzell (approved):

baltzell:
github wants Whitney to approve again, I think

baltzell (approved):

whit2333 (approved):

PR #1166, opened 2026-03-25
fix: don't let `SplitLogManager` do anything if it's not the manager

c-dilks:


c-dilks:
I was trying to figure out why the TOF calibration suite is so verbose, and suspected this fix would be involved; it turns out this has nothing to do with the issue.

I decided to open this draft PR anyway, in case we end up wanting/needing this.

PR #1177, opened 2026-04-01
fix: avert unwanted log level changes from upstream `JDBCProvider`

c-dilks:
Calling `JDBCProvider.connect()` destroys the log level of certain loggers, _viz._ `org.freehep.math.minuit`. Instead of attempting to fix upstream, we can save desired log levels and restore them; this is the fix we pretty much do downstream anyway (_e.g._, in `calcode`), where we typically just keep setting the log level we want.


c-dilks:
Not good enough for `calcode`, something else is changing the log level too. Worse yet, the point at which the log level change varies from test to test, and I suspect there may be some data race making it very difficult to track down exactly where the unwanted change happens.

PR #905, opened 2025-10-23
bigfeat:  detector changed hook

baltzell:

PR #1179, opened 2026-04-03
fix:  reduce bank access and MC checks in DC hit reading

baltzell:


tongtongcao (approved):

baltzell:
Raffaella pointed out this will conflict with RRNS #905, so we'll wait until that's merged.

PR #1205, opened 2026-04-11
feat:  add auto-generated enums for HIPO schema 

baltzell:


pom.xml
```diff
@@ -263,6 +263,30 @@
...
+        <artifactId>maven-antrun-plugin</artifactId>
+        <version>3.1.0</version>
+        <executions>
+          <execution>
+            <id>bank-schema-enums</id>
+            <phase>validate</phase>
+            <goals>
+              <goal>run</goal>
+            </goals>
+            <inherited>false</inherited> <!-- make this one NOT inherited -->
```
c-dilks: Instead of preventing inheritance, could you just move this plugin to `common-tools/clas-io`? There it would be closer to its generated Java code artifacts.

common-tools/clas-io/bank-schema-enums
c-dilks, line 1: If you move the plugin to `common-tools/clas-io`, you could also move this script to `common-tools/clas-io/` too; no need for it to live in `libexec` since this script is only needed at build time.

pom.xml
```diff
@@ -428,6 +452,22 @@
...
+              <fileset>
+                <directory>common-tools/clas-io/src/main/java/org/jlab/io/banks</directory>
+                <includes>
+                  <include>**/*.java</include>
+                </includes>
+              </fileset>
+            </filesets>
+          </configuration>
+        </plugin>
+
```
c-dilks: same here, this could be moved to `common-tools/clas-io`.

c-dilks (approved):

PR #1210, opened 2026-04-17
fix:  reduce bank access in DC engine

### LINKED ISSUE #1139: DC rereads start time from HIPO for every DC hit

baltzell:


baltzell:
github issues

PR #1215, opened 2026-04-18
feat:  more lookup optimizations in decoding

baltzell:
* Sets instead of Lists for lookup speed
* Minimize manual lookups for `/daq/fadc` tables

### DIFF
--- a/common-tools/clas-detector/src/main/java/org/jlab/detector/decode/DetectorEventDecoder.java
+++ b/common-tools/clas-detector/src/main/java/org/jlab/detector/decode/DetectorEventDecoder.java
@@ -26,16 +26,18 @@ public class DetectorEventDecoder {
 
     HashSet<DetectorType> keysFilter;
     HashSet<DetectorType> keysMicromega;
-    HashMap<DetectorType,String> translations;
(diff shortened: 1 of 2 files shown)


common-tools/clas-detector/src/main/java/org/jlab/detector/decode/DetectorEventDecoder.java
```diff
@@ -22,27 +24,22 @@ public class DetectorEventDecoder {
...
 
+    int runNumber = 10;
+ 
     public DetectorEventDecoder(boolean development){
-        if(development==true){
-            this.initDecoderDev();
-        } else {
-            this.initDecoder();
-        }
+        this.initDecoder();
```
raffaelladevita, line 42: should probably remove the development mode from the command line options

raffaelladevita (dismissed):

common-tools/clas-detector/src/main/java/org/jlab/detector/decode/DetectorEventDecoder.java
baltzell: Yeah, I thought it'd been defunct/notuseful for many years, right?

PR #1221, opened 2026-04-23
fix:  run-clara port-finding on macos

baltzell:
Based only on trial and error, higher port numbers seem to be necessary on Tahoe/macos26

### DIFF
--- a/bin/run-clara
+++ b/bin/run-clara
@@ -127,7 +127,7 @@ function get_host_ip() {
 }
 function get_dpe_port() {
     local ports
-    ports=$(seq 7000 20 8000)
+    ports=$(seq 8000 20 9000)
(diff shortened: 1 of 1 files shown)


bin/run-clara
```diff
@@ -127,7 +127,7 @@ function get_host_ip() {
 }
 function get_dpe_port() {
     local ports
-    ports=$(seq 7000 20 8000)
+    ports=$(seq 8000 20 9000)
```
c-dilks: I asked copilot about this, see #1228 

```suggestion
    ports=$(seq 49152 20 65535)
```

baltzell: confirmed to work

c-dilks (approved):

PR #1222, opened 2026-04-24
feat:  decouple macos/linux build jobs to reduce CI latency

baltzell:
Previously the downstream tests always wait on the macos build, which are much slower than the linux builds, but only actually use the linux builds.  This avoids that latency at the cost of some duplication ...

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -310,6 +310,36 @@ jobs:
           echo "COMMAND: ${{ matrix.cmd }}"
           ${{ matrix.cmd }}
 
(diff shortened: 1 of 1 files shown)


c-dilks (approved):

c-dilks:
@baltzell @raffaelladevita this PR requires a change in the list of CI checks which are required for merging. I will update the list; however, all older PRs will need to be updated with respect to the `development` branch so the new CI jobs wills run.

baltzell:
I removed the build jobs completely from the requirements, implicitly required by the test jobs anyway ...

PR #1243, opened 2026-05-04
fix:  reduce CVT atan calculations

baltzell:


baltzell:
after #1243, #1244, and #1245, the effect seems evident:

before:  https://userweb.jlab.org/~baltzell/tmp/a
after:  https://userweb.jlab.org/~baltzell/tmp/cpu.html

raffaelladevita (approved):

PR #1251, opened 2026-05-08
feat:  keep track of field types during initialization instead of checking later

baltzell:
It's a hotspot!  This supersedes the (previous) magfield component of #1236.


heddle (approved):
I approve of this change.

PR #1249, opened 2026-05-08
feat:  prioritize primary usage case for performance

baltzell:
`MagneticFields.hasActiveTorus/Solenoid` is a hot spot, similar to `CompositeField.hasTorus/Solenoid` in #1247

PR #1252, opened 2026-05-08
feat:  keep track of field types during initialization instead of checking later

baltzell:
`MagneticFields.hasActiveTorus/Solenoid` are hot spots.

This supersedes #1249, which only optimized for our most common use case.


heddle (commented):
I approve of this change.

heddle (approved):
I approve this change.

PR #1258, opened 2026-05-13
fix:  little DC speedups, finals for clarity

baltzell:


reconstruction/dc/src/main/java/org/jlab/rec/dc/cluster/ClusterCleanerUtilities.java
```diff
@@ -328,11 +323,8 @@ public FittedCluster LRAmbiguityResolver(DataEvent event, FittedCluster fClus,
         
         int index = 0;
         for(FittedHit hit: fClus) {
-   
             if (hit.get_Doca() < 0.4 * hit.get_CellSize()) {
                 hit.set_LeftRightAmb(0);
-            }
-            if (hit.get_LeftRightAmb() == 0) {
```
tongtongcao, line 335: The change might change index values.
For a hit with hit.get_Doca() >= 0.4, the original LR is 0, then index will not be added by 1 after change.

PR #1262, opened 2026-05-14
feat:  add event sorting to hipo-diff

baltzell:


common-tools/clas-io/src/main/java/org/jlab/utils/HipoDiff.java
```diff
@@ -157,22 +164,64 @@ public static void main(String args[]) {
...
+                if (m.containsKey(eventNumber)) {
+                    System.out.println("WARNING:  Duplicate event number " + eventNumber + " at event index " + eventIndex);
+                    duplicates.add(eventNumber);
+                }
+                else {
+                    m.put(eventNumber, eventIndex);
+                }
+            }
+        }
+        m.keySet().removeAll(duplicates);
```
raffaelladevita, line 191: it would be good to print a warning message with the event number

raffaelladevita (changes requested):

raffaelladevita (approved):

PR #1270, opened 2026-05-18
feat:  use previous unix time reading in CLARA writer

### LINKED ISSUE #1269: unixTime
Prior the recent #1263 , `RUN::config.unixTime` was being cached across events unless nonzero in EVIO, which was not nice for multi-threading the decoder.  _But now we likely have many events with `unixTime=0`, as that was apparently an old DAQ "fix"!_

We may want to revert to the previous non-reproducible situation, or reassign unix time in post-processing.

baltzell:
Prior to #1263, we were caching the unix time across events in decoding.  That was (presumably) very intentional, as the DAQ often writes zero for unix time, but it leads to non-determinism when parallelizing the decoder.

Here we instead write the "correct" unix time to each event in the CLARA writer.


baltzell:
still non-deterministic

PR #1273, opened 2026-05-19
feat:   hash sets/maps over lists in decoding

baltzell:
Just bookkeeping cleanup really, but it will enable removing another loop in #1323

### DIFF
--- a/common-tools/clas-detector/src/main/java/org/jlab/detector/decode/DetectorEventDecoder.java
+++ b/common-tools/clas-detector/src/main/java/org/jlab/detector/decode/DetectorEventDecoder.java
@@ -1,7 +1,6 @@
 package org.jlab.detector.decode;
 
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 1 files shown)


common-tools/clas-detector/src/main/java/org/jlab/detector/decode/DetectorEventDecoder.java
```diff
@@ -71,8 +71,11 @@ public void setVariation(String variation) {
     public void setRunNumber(int run){
         if (run != this.runNumber) {
             translator = new TranslationTable();
-            for (int i=0; i<keysTrans.size(); i++)
-                translator.add(keysTrans.get(i), translationManager.getConstants(run, tablesTrans.get(i)));
+            fitterTables = new HashMap<>();
+            for (DetectorType t : transTableNames.keySet())
```
raffaelladevita, line 75: What happens if a different run number is found and the add/put methods are used a second time with an already existing key?

baltzell: That's taken care of by the "new" objects a couple lines earlier.

raffaelladevita (approved):

PR #1280, opened 2026-05-20
feat:  cleanup io classes in yamls

baltzell:
Switch everything to `Clas12Reader/Writer` with `DecoderEngine` first in the chain.

### DIFF
--- a/etc/services/mc-ai.yaml
+++ b/etc/services/mc-ai.yaml
@@ -1,10 +1,10 @@
 io-services:
   reader:
-    class: org.jlab.io.clara.HipoToHipoReader
-    name: HipoToHipoReader
+    class: org.jlab.io.clara.Clas12Reader
+    name: Clas12Reader
   writer:
-    class: org.jlab.io.clara.HipoToHipoWriter
-    name: HipoToHipoWriter
+    class: org.jlab.io.clara.Clas12Writer
+    name: Clas12Writer
 services:
   - class: org.jlab.clas.swimtools.MagFieldsEngine
     name: MAGFIELDS
--- a/etc/services/mc-ai.yaml
+++ b/etc/services/mc-ai.yaml
@@ -6,6 +6,8 @@ io-services:
     class: org.jlab.io.clara.Clas12Writer
     name: Clas12Writer
 services:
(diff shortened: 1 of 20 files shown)


etc/services/mc-ai.yaml
```diff
@@ -1,11 +1,13 @@
...
-    name: HipoToHipoReader
+    class: org.jlab.io.clara.Clas12Reader
+    name: Clas12Reader
   writer:
-    class: org.jlab.io.clara.HipoToHipoWriter
-    name: HipoToHipoWriter
+    class: org.jlab.io.clara.Clas12Writer
+    name: Clas12Writer
 services:
+  - class: org.jlab.clas.reco.DecoderEngine
```
raffaelladevita: why adding the decoder to the mc yamls?

etc/services/swaps.yaml
```diff
@@ -1,13 +1,15 @@
...
-    name: HipoToHipoReader
+    class: org.jlab.io.clara.Clas12Reader
+    name: Clas12Reader
   writer:
-    class: org.jlab.io.clara.HipoToHipoWriter
-    name: HipoToHipoWriter
+    class: org.jlab.io.clara.Clas12Writer
+    name: Clas12Writer
 services:
+  - class: org.jlab.clas.reco.DecoderEngine
```
raffaelladevita: same question here

etc/services/mc-ai.yaml
baltzell: yeah, we don't have to

etc/services/swaps.yaml
baltzell: why not here?

raffaelladevita: because if we decode from evil, there is no reason to use the swap service

etc/services/mc-ai.yaml
baltzell: fixed

etc/services/swaps.yaml
baltzell: fixed

raffaelladevita (approved):

PR #1294, opened 2026-05-29
fix:  restore public access, used in dc calibration suite

baltzell:


c-dilks (approved):

c-dilks:
someday our CI should trigger consumers' CI, so we can catch stuff like this...

PR #1304, opened 2026-06-04
Iss 1297

N-Plx:


N-Plx:
ATOF calibrations to correct for the radiation damage effects that increases time offsets with time. It’s the slope that we see in these timelines:
 
image

This slope is corrected per blocks of runs, as it varies in different periods. A timing correction is defined as dt = (run number-reference run)*slope. This means that we need to know the run number for each hit to compute the full-time correction. Run number is read out at the engine level and propagated to the hits.

baltzell:
code.jlab.org had issues yesterday, hence the git-lfs errors in the CI jobs.  Should be resolved as of this morning.

N-Plx:
Zhiwan's validation of the timing corrections shows the offsets are now well corrected for:
image

mathieuouillon:
Looks good to me! The only thing I'd consider is building the key from `(sector, layer, component, order)` once and reusing it across the four `getDoubleValue` calls on lines 207/8 in reconstruction/alert/src/main/java/org/jlab/rec/atof/hit/ATOFHit.java (), since they all share the same keys. It's an optimization, though, so only worth it if this path is hot. So feel free to take it or leave it.

Minor nit: `Run` is capitalized for a field/parameter, which goes against the usual Java convention (`run`).

mathieuouillon (approved):

PR #1309, opened 2026-06-10
Clean bad AHDC hits and use ATOF wedges in the Kalman Filter

ftouchte:
Features:
- Make the Kalman Filter reusable (store error covariance matrix in Track)
- Add a routine to clean bad AHDC hits
- Use ATOF wedge in the Kalman Filter
- Extends` AHDC::kftrack`
- Convenient use of the PID in RungeKutta.java

Code consistency:
- Create an interface for the hits used by the Kalman Filter : `KFHit`
   - AHDC hits are stored in the usual `Hit`
   -  beamline, ATOF wedges/bars are stored in `RadialKFHit`
   - the method `KFitter.correct()` no longer needs to explicitly check if we works with an AHDC hits or not: the polymorphism do the job for use
- Remove redundancy : `KFitter` does not use another `Stepper` in its constructor (the initial `Stepper` and the initial state are linked)
   
This pull request is the sync version of this draft : https://github.com/JeffersonLab/coatjava/pull/1117

**In coatjava development**
image
**This PR**
image


etc/bankdefs/hipo4/alert.json
```diff
@@ -408,6 +408,74 @@
         "name": "sum_residuals",
         "type": "F",
         "info": "Sum of residuals (mm)"
+      }, {
```
mathieuouillon, line 411: It starts to be a huge bank. Is it useful to have all this debugging information in the main database?

reconstruction/alert/src/main/java/org/jlab/service/alert/ALERTEngine.java
```diff
@@ -562,24 +601,156 @@ public boolean processDataEventUser(DataEvent event) {
...
+
+        /// Clean AHDC bad hits
+        double sigma = 0.5; // mm
+        for (Track track : AHDC_tracks) {
+            ArrayList<Hit> AHDC_hits = track.getHits();
+            Iterator<Hit> it = AHDC_hits.iterator();
+            while (it.hasNext()) {
+                Hit hit = it.next();
+                if (Math.abs(hit.getResidual()) > 3*sigma) {
+                    it.remove();
```
mathieuouillon, line 739: What happens to a track which have less than AHDC 3 hits after the cleanup? The chi2/ndf will be infinite.

reconstruction/alert/src/main/java/org/jlab/rec/ahdc/KalmanFilter/KalmanFilter.java
```diff
@@ -31,149 +32,379 @@ public class KalmanFilter {
...
-                    chi2 += Math.pow(hit.getResidual(),2)/hit.get_MeasurementNoise().getEntry(0,0);
+                    chi2 += Math.pow(hit.getResidual(),2)/hit.getMeasurementNoiseMatrix().getEntry(0,0);
 			    }
-			    track.set_sum_adc(sum_adc);
 			    track.set_sum_residuals(sum_residuals);
 			    track.set_chi2(chi2/(AHDC_hits.size()-3));
 			    track.set_p_drift(p_drift);
-			    track.set_dEdx(sum_adc/s);
 			    track.set_path(s);
-			    track.set_n_hits(AHDC_hits.size());
```
mathieuouillon, line 170: n_hits and sum_adc are no longer set. It should be fix, or these variables can be removed from the bank

reconstruction/alert/src/main/java/org/jlab/rec/ahdc/KalmanFilter/KalmanFilter.java
```diff
@@ -31,149 +32,379 @@ public class KalmanFilter {
...
+						track.set_ATOF_region(3); // indicate if we reach this surface or not
+					}
+
+				} // end propagation towards ATOF surface
+			    
 			}//end of loop on track candidates
 		} catch (Exception e) {
-			//e.printStackTrace();
-			//System.out.println("======> Kalman Filter Error");
+			e.printStackTrace();
```
mathieuouillon: Use the logger, probably not a good Idea to print the full Trace in production/cooking

reconstruction/alert/src/main/java/org/jlab/rec/ahdc/KalmanFilter/KFitter.java
```diff
@@ -86,32 +84,12 @@ public void predict(Hit hit, boolean direction) throws Exception {
...
-		} 
-		// else, it is an AHDC hits
-		else {
-            measurementNoise = hit.get_MeasurementNoise();//1x1
-            measurementMatrix = H(stateEstimation, hit);//6x1
-            h = h(stateEstimation, hit);//1x1
-			z = hit.get_Vector();//1x1
-		}
+	public void correct(KFHit hit) {
+        RealVector z = hit.getMeasurementVector();
```
mathieuouillon: Not sure where to leave that comment. 
With hits represented as r, phi, z, an angle issue can arise. The second component of RadialKFHit is φ = atan2(y, x). There is no ±π wrapping. For an ATOF wedge whose azimuth sits near the branch cut, the measured and predicted φ can differ by ≈2π even when physically identical, producing a huge bogus innovation that corrupts the update. 
Have you taken care of that?

ftouchte:
Hello everyone, this pull request is ready for review.

I'm waiting for this PR to get validated before submitting the next one about the alignment: https://github.com/JeffersonLab/coatjava/pull/1310

mathieuouillon:
I've added a few comments. Sorry, I forgot to send them.

etc/bankdefs/hipo4/alert.json
ftouchte: It's on demand of Raphael. In the final version of the code, once we will agree what should be kept or not, there will be a lot of cleaning.

reconstruction/alert/src/main/java/org/jlab/service/alert/ALERTEngine.java
ftouchte: In the analysis, I usually have cut on the number of hits (>= 6 or 7) to select good tracks. 

The chi2/ndf will be infinite or negative if we have less then 3 hits but we will know that are not good tracks. 

In any case, at the Kalman Filter level, we should not have tracks with less than 5 hits. We still need to decide and document the exact lower limit of the number of hits, so that it can be implemented in the TrackFinder for example.

etc/bankdefs/hipo4/alert.json
mathieuouillon: Ok

reconstruction/alert/src/main/java/org/jlab/rec/ahdc/KalmanFilter/KFitter.java
ftouchte: That is interesting. I see what you mean, the issue is here

`RealVector innovation = z.subtract(h);`

We expect $\phi_z - \phi_h$ to be close to zero, but by the way they are computed, they can be close to 0 and $2 \pi$. 

They should represent a small fraction of the events but I haven't taken care of that. 

I have an idea to fix it. It will be fixed in the following commits.

reconstruction/alert/src/main/java/org/jlab/rec/ahdc/KalmanFilter/KalmanFilter.java
ftouchte: Fixed by this commit : https://github.com/JeffersonLab/coatjava/pull/1309/commits/2e8d72294a69b2fc1807516000098c23970a1dd2

reconstruction/alert/src/main/java/org/jlab/rec/ahdc/KalmanFilter/KalmanFilter.java
ftouchte: Fixed by this commit : https://github.com/JeffersonLab/coatjava/pull/1309/commits/0f2c8542c692188d03e5f1814575de875eeea617

reconstruction/alert/src/main/java/org/jlab/rec/ahdc/KalmanFilter/KFitter.java
ftouchte: Fixed by this commit: https://github.com/JeffersonLab/coatjava/pull/1309/commits/804a9647c22f1bcd84ab4d3707787d1943ee15bc

reconstruction/alert/src/main/java/org/jlab/rec/ahdc/KalmanFilter/KalmanFilter.java
mathieuouillon: The variables would still exist in the bank, no?

ftouchte: In AHDC::kftrack, yes. As it is, they will probably differ from the values in AHDC::track, and that's expected.

ftouchte: We no longer need to set them as they are automatically computed by the getter.

mathieuouillon (dismissed):

reconstruction/alert/src/main/java/org/jlab/service/ahdc/AHDCEngine.java
```diff
@@ -96,6 +96,7 @@ public boolean processDataEventUser(DataEvent event) {
             ArrayList<Hit> AHDC_Hits = hitReader.get_AHDCHits();
 
             RecoBankWriter writer = new RecoBankWriter();
+            event.removeBank("AHDC::hits");
```
baltzell: Should registerOutputs method should be used instead?  Its purpose is to remove output banks, so the engines don't have to.

ftouchte: Fixed by this commit : https://github.com/JeffersonLab/coatjava/pull/1309/commits/b3770d613dbcbcf779f67a6f46055f7d1319db6c

mathieuouillon (approved):

PR #1316, opened 2026-06-15
remove unused imports and variables, run auto-indenter

baltzell:


silvianic:
I think everything is ok

PR #1319, opened 2026-06-17
example:  using maven repos from code.jlab.org

baltzell:
And see https://code.jlab.org/hallb/clas12/coatjava/clas-physics for how that maven repo was CI-made.


baltzell:
In use for hipo-java

PR #1328, opened 2026-06-30
fix macos CI test, add SQLites to coatjava build

baltzell:
* Just remove reliance on CVMFS from github CI
* Add https://code.jlab.org/hallb/clas12/ccdbrcdb-sqlite for git-lfs storage of CCDB/RCDB SQLite files
* Add SQLite to the build script defaults (another 500 MB), add `--nosqlite` option

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,12 +66,6 @@ jobs:
           java-version: ${{ env.JAVA_VERSION }}
           distribution: ${{ env.java_distribution }}
           cache: maven
-      - name: setup cvmfs
-        uses: cvmfs-contrib/github-action-cvmfs@v5
-        with:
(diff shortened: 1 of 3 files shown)


c-dilks (commented):
`build-coatjava.sh` now downloads the SQLite files by default, which could be out of date with respect to the "official" databases. However, these downloaded SQLite files are _only_ used if `$CCDB_CONNECTION` or `$RCDB_CONNECTION` point to them (as you do in the CI here); otherwise, the databases that are used by default remain unchanged.

Do I understand that correctly?

c-dilks:
Could we have a more descriptive PR title (for the release notes)?

baltzell:
Yes, the SQLite snapshots are now included by default, but only used if one figures out how to set $CCDB/RCDB_CONNECTION to point to them.

c-dilks (approved):

PR #1329, opened 2026-06-30
fix: `QadbBin.getEventNum{Min,Max}` return value type

c-dilks:


baltzell:
proposal to remove cvmfs from github ci, too unreliable:  https://github.com/JeffersonLab/coatjava/pull/1328

baltzell (approved):

PR #1331, opened 2026-07-01
ci: update `needs` for `final` job

c-dilks:
with this, we could set `final` to be a required job to pass for CI pipelines


baltzell (approved):

c-dilks:
@baltzell I added `final` to the list of required CI status checks in the `development` branch protection rules.

PR #1338, opened 2026-07-08
feat(QCDDAT): CVT ML Sub-package

c-dilks:
PR opened on behalf of @zieglerv

Redux of `CVT_QCDDAT` branch, since
- keep @zieglerv's original branch as is
- branch `urwt_detector-change-hook` is based off of `CVT_QCDDAT`


c-dilks:
Hmm.. Look's like Veronique's commits are authored by `ziegler` (which GitHub thinks is an unknown user), not @zieglerv; anyway, it's clear these are both Veronique, but we should make sure @zieglerv is attributed when we merge this (otherwise it may erroneously look like I did all this work, when all I've done so far was rebase!)

zieglerv:
Sorry I committed them from my personal Mac that is why the author name is different

baltzell:
See #1339 to address the ReconstructionEngine issues.

PR #1349, opened 2026-07-14
remove debug printouts and timing stuff

baltzell:
depends on #1348


c-dilks:
@raffaelladevita and @baltzell, here is the HEP data model for PID:

https://github.com/key4hep/EDM4hep/blob/8aa0fcd2176bc1981bd73b7558bc94d0cbcf6453/edm4hep.yaml#L380-L391

In the EIC use case, we have a 1-to-many relation from a reconstructed particle object to `edm4hep::ParticleID` objects, and a 1-to-1 relation to the "best" one; the "best" `edm4hep::ParticleID` object's `PDG` was assumed to be that reconstructed particle's PDG for anything that needs it downstream, _e.g._, kinematics calculations ($x$, $Q^2$, etc.).

For our case at CLAS12, I think the `pindex` column gives our "1-to-many" mapping of `REC::Particle` to `REC::Hypothesis` objects (see below), but I think we also need at least:
- a `status`-_like_ variable, _not_ `REC::Particle.status`, would be useful for _each_ hypothesis, indicating what was involved in that PID hypothesis; it would encode which detector(s), which algorithm(s), etc.
- Some way of representing _which_ `REC::Hypothesis` was assumed for setting the `pid` column of the `REC::Particle`

This is in reference to the schema proposed in this PR, copied here:
```json
{
        "name": "REC::Hypothesis",
        "group": 300,
        "item": 50,
        "info": "Particle identification bank",
        "entries": [
            {"name":"pindex","type":"S", "info":"row index in REC::Particle"},
            {"name":"pid",   "type":"I", "info":"particle id hypothesis in LUND conventions"},
            {"name":"logl",  "type":"F", "info":"log(Likelihood) for this hypothesis"}
        ]
}
```

Issue #1336, opened 2026-07-07
DC combo engine errors


baltzell:
```
│ java.lang.RuntimeException: Batch prediction error in 6ClsComboClf                                                                │
│         at org.jlab.service.ai.DCClsComboEngine.processDataEvent(DCClsComboEngine.java:183)                                       │
│         at org.jlab.clas.reco.ReconstructionEngine.filterEvent(ReconstructionEngine.java:368)                                     │
│         at org.jlab.clas.reco.ReconstructionEngine.execute(ReconstructionEngine.java:407)                                         │
│         at org.jlab.clara.sys.ServiceEngine.executeEngine(ServiceEngine.java:227)                                                 │
│         at org.jlab.clara.sys.ServiceEngine.execute(ServiceEngine.java:153)                                                       │
│         at org.jlab.clara.sys.Service.lambda$execute$2(Service.java:178)                                                          │
│         at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:572)                                      │
│         at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)                                                     │
│         at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)                              │
│         at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)                              │
│         at java.base/java.lang.Thread.run(Thread.java:1583)                                                                       │
│ Caused by: ai.djl.translate.TranslateException: java.lang.IllegalArgumentException: capacity < 0: (-1996400944 < 0)               │
│         at ai.djl.inference.Predictor.batchPredict(Predictor.java:200)                                                            │
│         at ai.djl.inference.Predictor.predict(Predictor.java:133)                                                                 │
│         at org.jlab.service.ai.DCClsComboEngine.processDataEvent(DCClsComboEngine.java:177)                                       │
│         ... 10 more                                                                                                               │
│ Caused by: java.lang.IllegalArgumentException: capacity < 0: (-1996400944 < 0)                                                    │
│         at java.base/java.nio.Buffer.createCapacityException(Buffer.java:290)                                                     │
│         at java.base/java.nio.Buffer.<init>(Buffer.java:253)                                                                      │
│         at java.base/java.nio.ByteBuffer.<init>(ByteBuffer.java:316)                                                              │
[...]

PR #1352, opened 2026-07-16
update DCClsComboEngine.java to set limit for cluster combos in case of memory explosion

tongtongcao:
Response to #1336,  set limit for DC cluster combos. 
By default, limit is 1M for each sector. If combos reach the limit for a sector, then the first 1M combos will be kept.
Totally, the memory usage is estimate to be ~1.8G at most, if all 6 sectors reach the default limit.
Limit of 1M is supposed to be large enough. The plot show distribution for # of combos of a sector using run 6666 for test.
Plot_07-16-2026_09 43 03_AM

### DIFF
--- a/reconstruction/ai/src/main/java/org/jlab/service/ai/DCClsComboEngine.java
+++ b/reconstruction/ai/src/main/java/org/jlab/service/ai/DCClsComboEngine.java
@@ -57,6 +57,10 @@ public class DCClsComboEngine extends ReconstructionEngine {
             
     final static int SUPERLAYERS = 6;    
     
+    private static final Logger LOGGER = Logger.getLogger(DCClsComboEngine.class.getName());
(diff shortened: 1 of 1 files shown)


baltzell:
Isn't 1e6 very conservative?   Your data stops at 8e4?

baltzell:
And what's the max memory needed for 1e6?

tongtongcao:
The test is after denoising, and run 6666 is 50nA. Consider much higher luminosity for the future and potential reconstruction without denoising, I set much higher value for limit. 
I think that 1.8G memory requirement at most is acceptable.
What's your suggestion?

tongtongcao:
Reset default limit of cluster combos for a sector from 1M to 100k.

baltzell (approved):

PR #1356, opened 2026-07-21
build: remove `javafx` from modules that don't really need it

c-dilks:
Now `javafx` is only consumed by `common-tools` stuff:
```bash
$ rg javafx -gpom.xml -g!/pom.xml -l
```
```
common-tools/clas-geometry/pom.xml
common-tools/clas-jcsg/pom.xml
```


baltzell:
java still has no Pair:)

baltzell (approved):

PR #1357, opened 2026-07-22
Rich clustering update

mirazita:
Fix on the clustering algorithm.
Added check on the same sector when making clusters.


baltzell:
Also sorts clusters now.

raffaelladevita (approved):

PR #1366, opened 2026-07-27
feat: CD denoising

c-dilks:
Rebased from @zieglerv's `cd_denoising` branch, after removing the HIPO-to-NPZ stuff.

#1365 is vice-versa: `cd_denoising` without the CD denoising stuff.


baltzell (approved):

tongtongcao:
The updates are for application of AI denoising model for CVT hits. Still, there will be more updates for removal of CVT hits based on AI prediction. 
We will use the branch for the further updates, so I do not suggest to merge the branch before all denoising-related updates are done.

PR #1368, opened 2026-07-30
feat: ALERT pre- and post-KF AI PID inference

skuditha:
- Replaces the existing ALERT pre-PID implementation with the new AHDC- and ATOF-based TorchScript models.
- Adds post-KF PID inference for every valid ALERT::ai:projections pair.
- Writes predicted particle IDs and class probabilities to the existing ALERT AI PID banks.

### DIFF
--- a/reconstruction/alert/src/main/java/org/jlab/service/alert/ALERTEngine.java
+++ b/reconstruction/alert/src/main/java/org/jlab/service/alert/ALERTEngine.java
@@ -820,11 +820,11 @@ public class ALERTEngine extends ReconstructionEngine {
                         LOGGER.warning(() -> "Exception in ALERTEngine PostPID: " + ex);
                     }
                 }
-				rbc.appendPIDBank(event, pidResults);
             }
-            
+                if (pidResults != null && !pidResults.isEmpty()) {
+                    rbc.appendPIDBank(event, pidResults);
+            }     
         }
-
         return true;
     }
 
--- a/reconstruction/alert/src/main/java/org/jlab/service/alert/ALERTEngine.java
+++ b/reconstruction/alert/src/main/java/org/jlab/service/alert/ALERTEngine.java
@@ -24,6 +24,7 @@ import org.jlab.io.hipo.HipoDataSource;
 import org.jlab.io.hipo.HipoDataSync;
 import org.jlab.rec.alert.TrackMatchingAI.ModelTrackMatching;
 import org.jlab.rec.alert.AIPID.ModelPrePID;
+import org.jlab.rec.alert.AIPID.ModelPostPID;
 import org.jlab.rec.alert.banks.RecoBankWriter;
 import org.jlab.rec.alert.projections.TrackProjector;
 import org.jlab.rec.atof.hit.ATOFHit;
@@ -60,6 +61,7 @@ import java.util.logging.Logger;
 
 import ai.djl.util.Pair;
 import org.jlab.rec.alert.AIPID.PrePIDResult;
+import org.jlab.rec.alert.AIPID.PIDResult;
 
 
 /**
@@ -115,6 +117,7 @@ public class ALERTEngine extends ReconstructionEngine {
 
     private ModelTrackMatching modelTrackMatching;
     private ModelPrePID modelPrePID;
+    private ModelPostPID modelPostPID;
 
     // AHDC track-finding strategy (driven by ALERT.Mode YAML key)
     private TrackFinder trackFinder;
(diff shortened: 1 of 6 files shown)


reconstruction/alert/src/main/java/org/jlab/service/alert/ALERTEngine.java
```diff
@@ -178,7 +182,8 @@ public boolean init() {
                 "AHDC::preclusters", "AHDC::clusters", "AHDC::track",
                 "AHDC::interclusters", "AHDC::docaclusters", "AHDC::ai:prediction",
                 "AHDC::mc", "AHDC::kftrack",
-                "ALERT::projections", "ALERT::ai:projections", "ALERT::prePID");
+                "ALERT::projections", "ALERT::ai:projections", 
+                "ALERT::ai:prePID", "ALERT::ai:postPID");
```
mathieuouillon: Wrong banks name: look into etc/bankdefs/hipo4/alert.json
"name": "ALERT::ai:prepid", and name": "ALERT::ai:pid"

reconstruction/alert/src/main/java/org/jlab/service/alert/ALERTEngine.java
```diff
@@ -766,11 +761,105 @@ public boolean processDataEventUser(DataEvent event) {
...
+                        float[] prediction = modelPostPID.prediction(features);
+                        if (prediction != null) {
+                            pidResults.add(new PIDResult(trackid, clusterid, prediction));
+                        }
+                    } catch (TranslateException ex) {
+                        LOGGER.warning(() -> "Exception in ALERTEngine PostPID: " + ex);
+                    }
+                }
+            }
+            rbc.appendPIDBank(event, pidResults);
```
mathieuouillon: appendPIDBank is outside the isFinite(startTime) guard, so a zero-row ALERT::ai:pid is appended to every qualifying event. Consider if (!pidResults.isEmpty()).

mathieuouillon (commented):
The old translator did logits.softmax(1) before toFloatArray(). Both new translators (ModelPrePID.java:290, ModelPostPID.java:84) read the raw output tensor. argmax is unaffected, so pid/prepid are fine either way, but if the exported TorchScript modules end in a Linear rather than a Softmax, then p2212/prob_2212/... are logits, and downstream users of a bank field literally named prob_* will be misled. Can you confirm the export includes the softmax?

reconstruction/alert/src/main/java/org/jlab/service/alert/ALERTEngine.java
skuditha: (1) Confirmed. I inspected the TorchScript graphs for both pre-PID models and the post-KF model. All three exported modules apply `softmax(..., -1)` inside their `forward()` method before returning the output tensor. Therefore, the values read by the new translators are already normalized class probabilities, not logits. Applying another softmax in Java would double-transform the probabilities, so the current translators are correct as written.
(2) Thank you! I will fix it. But it does not affect the output for some reason.
(3) Will do. Again, a zero row ALERT::ai:pid did not appear in my tests. So, there seems to be no effect.

skuditha:
Resolved Mathieu's comments. CI will fail until the pull request to merge the models to the networks repository is approved. (https://code.jlab.org/hallb/clas12/coatjava/networks/-/merge_requests/13)

reconstruction/alert/src/main/java/org/jlab/service/alert/ALERTEngine.java
skuditha: For (3) The reason why the zero row banks don't appear is because there is another guard in the recobankwriter. Just to be safe, I pulled the appendPIDBank inside the starttime guard.

c-dilks:
I merged https://code.jlab.org/hallb/clas12/coatjava/networks/-/merge_requests/13 and manually triggered @dependabot to create https://github.com/JeffersonLab/coatjava/pull/1369.
Once that's merged, you can rebase this branch or merge `development` into it.

baltzell:
@skuditha Are you sure you addressed all?  Looks like the bank names may still not exist.  (Also, you can mark them "resolved" if they are so.)

reconstruction/alert/src/main/java/org/jlab/service/alert/ALERTEngine.java
skuditha: added additional guard

baltzell (approved):

PR #1373, opened 2026-08-03
perf: `hipo2npz` memory efficiency

c-dilks:
The current `hipo2npz` is not memory efficient, since it effectively reads and stores a full HIPO files' data structures in memory while producing the NPZ file.

This PR replaces the memory storage with disk storage, by creating a directory of temporary files in `<output>.tmp/`, where `<output>` is the output file name. Now the memory usage hovers under 500 MB typically for a DST file.

I had Claude generate an NPZ `diff` tool, stored here as `hipo2npz-diff`; I used it to verify there is no difference in the resulting NPZ file before and after this PR.

### About Temporary Files
- We do not use the `/tmp` file system, to keep the temp files and output NPZ file on the same disk.
- The temporary files are deleted when:
  - the conversion is done
  - anything that "nicely" kills the job, such as `SIGINT` or `SIGTERM`
- The temporary files remain if the job is not-so-nicely killed (`SIGKILL`), since the JVM does not have a chance to remove them in that case; this means if we run on SLURM, it may be possible that OOM or time-limit kills from SLURM will leave temp files.
- If a directory, file, or symlink with the name `<output>.tmp` already exists, `hipo2npz` will immediately fail, preventing accidental deletion of something that was not _created_ by `hipo2npz`

### Additional changes
- new options `--num-events` and `--first-event` to convert a range of events


etc/data/nnet
c-dilks, line 1: Oops. I need to revert this when the host is back online.

baltzell (approved):

PR #1374, opened 2026-08-04
feat: build option `--no-aux` to disable downloading of all auxiliary data

c-dilks:
I also renamed `--nomaps` -> `--no-maps`, _etc._; the old names still work as legacy aliases, so no changes are needed downstream.


c-dilks:
I'm glad we still have macOS CI testing here, since it ships with an ancient version of `bash`

baltzell (approved):

PR #1378, opened 2026-08-11
add occupancy bookkeeping class and HIPO banks

baltzell:
To reduce data volume, this adds an occupancy bookkeeping class and engine, with I/O helpers for indexed banks ready for all standard clas12 detectors.   The new occupancy banks currently look like this:
```
        "name": "OCC::FTOF::adc",
        "group": 500,
        "item": 14,
        "info": "Occupancy Bank",
        "entries": [
            { "name":"sector"    , "type":"B", "info":"sector"},
            { "name":"layer"     , "type":"B", "info":"layer"},
            { "name":"component" , "type":"S", "info":"component"},
            { "name":"hits"      , "type":"F", "info":"number of hits"}
```
- [ ] Choose occupancy bank naming.
- [x] Need to use `FilteredBank` for order-hijacking by denoising
- [ ] Check the #-index assumption for all detectors
- [ ] Missing detectors
- [x] Note, this ignore negative indices, e.g. BMT/BST
- [ ] YAML variable:  "prescale"?

The main consumer of these new occupancy banks is [clas12-timeline](https://github.com/jeffersonlab/clas12-timeline).

Depends on ~~#1380, #1381, #1382, #1383~~.


etc/bankdefs/hipo4/occupancy.json
```diff
@@ -0,0 +1,242 @@
...
+    {
+        "name": "OCC::DC::tdc",
+        "group": 500,
+        "item": 10,
+        "info": "Occupancy Bank",
+        "entries": [
+            { "name":"sector"    , "type":"B", "info":"sector"},
+            { "name":"layer"     , "type":"B", "info":"layer"},
+            { "name":"component" , "type":"S", "info":"component"},
+            { "name":"hits"      , "type":"F", "info":"number of hits"}
```
c-dilks, line 11: why `float` and not `int`?

common-tools/clas-detector/src/main/java/org/jlab/detector/calib/utils/OccupanceTable.java
```diff
@@ -0,0 +1,236 @@
...
+                new OccupanceTable(schema,"LTCC::adc"),
+                new OccupanceTable(schema,"LTCC::tdc"),
+                new OccupanceTable(schema,"BST::adc"),
+                new OccupanceTable(schema,"BMT::adc"),
+                new OccupanceTable(schema,"FTCAL::adc"),
+                new OccupanceTable(schema,"FTHODO::adc"),
+                new OccupanceTable(schema,"FTTRK::adc"),
+                new OccupanceTable(schema,"RICH::tdc"),
+                new OccupanceTable(schema,"BAND::adc"),
+                new OccupanceTable(schema,"BAND::tdc"),
```
c-dilks: Is there any way to avoid the DRY violation here, and read `occupancy.json` somehow?

etc/bankdefs/hipo4/occupancy.json
baltzell: so ~~because~~ it can be normalized

common-tools/clas-detector/src/main/java/org/jlab/detector/calib/utils/OccupanceTable.java
baltzell: That could surely be done.

baltzell:
Negative indices definitely doesn't work:

```
java.lang.IllegalArgumentException: Index value out of range (0–65535) : -119
	at org.jlab.utils.groups.IndexedList.validateIndex(IndexedList.java:86)
	at org.jlab.utils.groups.IndexedList.add(IndexedList.java:99)
	at org.jlab.utils.groups.IndexedTable.addEntry(IndexedTable.java:95)
	at org.jlab.detector.calib.utils.OccupanceTable.fill(OccupanceTable.java:82)
```

Probably this means the `addEntry` isn't being used for translation tables in the decoder (there's a string-based alternative) ....

baltzell:
Nope, IndexedTables in decoding always have positive indices (they're defined only by CCDB tables), and `addEntryFromString` has the same issue with negatives.  The negative is created on bank writing.

Note, the swaps service will throw an exception for RICH's `layer>128` hits.

c-dilks:
Some info about `group` and `item` here on pp. 11 ff. of Mathieu's HIPO spec writeup:

https://code.jlab.org/ouillon/hipo4j/-/raw/691447f5a212a23a9561055cb9985e104828794f/specification/hipo-file-format.pdf?inline=false

(CC @raffaelladevita)

common-tools/clas-detector/src/main/java/org/jlab/detector/calib/utils/OccupanceTable.java
baltzell: done!

common-tools/clas-detector/src/main/java/org/jlab/detector/calib/utils/OccupanceTable.java
```diff
@@ -0,0 +1,123 @@
...
+     * @param index 
+     */
+    public synchronized final void fill(float weight, int... index) {
+        for (int i=0; i<index.length; i++) if (index[i] < 0) return;
+        final long hash = IndexedTable.DEFAULT_GENERATOR.hashCode(index);
+        if (!table.hasEntryByHash(hash)) {
+            table.addEntry(index);
+            table.setDoubleValueByHash(0.0d, 0, hash);
+        }
+        table.setDoubleValueByHash(table.getDoubleValueByHash(0, hash) + weight, 0, hash);
```
raffaelladevita, line 79: should this be *weight?

etc/bankdefs/hipo4/occupancy.json
```diff
@@ -0,0 +1,242 @@
...
+    },
+    {
+        "name": "OCC::FTOF::adc",
+        "group": 500,
+        "item": 14,
+        "info": "Occupancy Bank",
+        "entries": [
+            { "name":"sector"    , "type":"B", "info":"sector"},
+            { "name":"layer"     , "type":"B", "info":"layer"},
+            { "name":"component" , "type":"S", "info":"component"},
```
raffaelladevita, line 58: do we want to keep order for FTOF, CTOF, CND, ...?

etc/bankdefs/hipo4/occupancy.json
```diff
@@ -0,0 +1,242 @@
...
+        "item": 29,
+        "info": "Occupancy Bank",
+        "entries": [
+            { "name":"sector"    , "type":"B", "info":"sector"},
+            { "name":"layer"     , "type":"B", "info":"layer"},
+            { "name":"component" , "type":"S", "info":"component"},
+            { "name":"hits"      , "type":"F", "info":"number of hits"}
+        ]
+    }
+]
```
raffaelladevita, line 266: CND is missing

reconstruction/calib/src/main/java/org/jlab/calibration/service/OccupanceEngine.java
```diff
@@ -0,0 +1,68 @@
...
+    OccupanceTable[] tables;
+        
+    public OccupanceEngine() {
+        super("Occupance", "baltzell","0.1");
+    }
+
+    @Override
+    public boolean processDataEventUser(DataEvent event) {
+        for (OccupanceTable t : tables) {
+            RawDataBank b = new RawDataBank(t.getHitBank(), 1000, OrderGroups.NOMINAL);
```
raffaelladevita: I think we may want NODENOISE instead of NOMINAL

reconstruction/calib/src/main/java/org/jlab/calibration/service/OccupanceEngine.java
```diff
@@ -0,0 +1,68 @@
...
+            b.read(event);
+            t.fill(b, false);
+        }
+        if (++events % prescale == 0) {
```
raffaelladevita: is this thread safe?

common-tools/clas-detector/src/main/java/org/jlab/detector/calib/utils/OccupanceTable.java
```diff
@@ -0,0 +1,123 @@
...
+        }
+    }
+
+    /**
+     * Get an occupancy bank, normalized by number of events.
+     * @param events
+     * @param event
+     * @return 
+     */
+    public DataBank create(long events, DataEvent event) {
```
raffaelladevita, line 108: in multithreading, is it possible the table gets modified by thread X while thread Y is creating the bank?

common-tools/clas-detector/src/main/java/org/jlab/detector/calib/utils/OccupanceTable.java
baltzell: No, `weight` is the increment here, 1 for normal counting.

baltzell: Maybe should change the variable name ...

reconstruction/calib/src/main/java/org/jlab/calibration/service/OccupanceEngine.java
baltzell: done

reconstruction/calib/src/main/java/org/jlab/calibration/service/OccupanceEngine.java
baltzell: done

etc/bankdefs/hipo4/occupancy.json
baltzell: done

PR #1388, opened 2026-08-18
make recon-util multi-threaded

baltzell:
Depends on #1389, #1390, #1391.


c-dilks:
how's the performance compare to that of `clara`, less overhead?

baltzell:
Not sure, we'll see soon ...

